PR #42, opened 2017-02-17
V8 refactored internals to support sync/async out of the box

dendril:


src/engine/combiners/and.js
```diff
@@ -17,27 +17,37 @@ limitations under the License.
...
 
-function andCombinerContext(matchers: Array<Function>): Function {
+function andResults(results) {
+  let i = 0;
+  let len = results.length;
+  let hasMatchedAll;
+
+  // loop through all the matchers an stop at the first one returning false.
+  for (; i < len && results[i]; i++) {
+    // logic is run inside the condition of evaluates next step.
```
patricioe: did you miss the content of this for loop?

src/engine/combiners/ifelseif.js
```diff
@@ -13,36 +13,50 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-    log('all predicates evaluted, none treatment available');
+  log('all predicates evaluted, none treatment available');
+  return undefined;
+}
+
+function ifElseIfCombinerContext(predicates: Array<Function>): Function {
+
+  function ifElseIfCombiner(key: SplitKey, seed: number, attributes: Object): Promise<?Evaluation> | ?Evaluation {
+    // In Async environments we are going to have async predicates. There is none way to know
```
patricioe: typo: "There is none way" should be "There is no way"

src/engine/combiners/and.js
dendril: Given a boolean array [ true, true, true, ... ] I'm computing true if and only if all the values are true.

PR #56, opened 2017-04-05
[SPLIT-2844] Addin try-catch block for key parsing

NicoZelaya:
We should discuss if there is a need for checking the type of the two properties, in case they are present on the key object.
Also, if we should notify customer in some way of the exception reason (maybe a console.error() ?)

[SPLIT-2844](https://splitio.atlassian.net/browse/SPLIT-2844)

### DIFF
--- a/src/engine/evaluator/index.js
+++ b/src/engine/evaluator/index.js
@@ -20,6 +20,7 @@ limitations under the License.
 
 const engine = require('../engine');
 const thenable = require('../../utils/promise/thenable');
+const LabelsConstants = require('../../utils/labels');
 
 // Build Evaluation object if and only if matchingResult is true
 function match(matchingResult: boolean, bucketingKey: string, seed: number, treatments: Treatments, label: string, algo: ?number): ?Evaluation {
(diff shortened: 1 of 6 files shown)


src/engine/evaluator/index.js
```diff
@@ -41,26 +40,21 @@ function match(matchingResult: boolean, bucketingKey: string, seed: number, trea
...
     }
 
     // matcherEvaluator could be Async, this relays on matchers return value, so we need
     // to verify for thenable before play with the result
-    const matches = matcherEvaluator(matchingKey, attributes);
+    const matches = matcherEvaluator(key.matchingKey, attributes);
 
     if (thenable(matches)) {
-      return matches.then(result => match(result, bucketingKey, seed, treatments, label, algo));
+      return matches.then(result => match(result, key.bucketingKey, seed, treatments, label, algo));
```
dendril: Lets review this because matching mechanism should use always the `matchingKey`, `bucketingKey` should be used only in the bucket calculation.

PR #65, opened 2017-04-17
Adding support for String Matchers

NicoZelaya:
Matcher types:
* Starts With
* Ends With
* Contains

### DIFF
--- a/src/engine/matchers/cont_all.js
+++ b/src/engine/matchers/cont_all.js
@@ -13,24 +13,22 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 **/
-'use strict';
 
-/*eslint-disable eqeqeq */
+'use strict';
 
 const log = require('debug')('splitio-engine:matcher');
 const intersection = require('lodash/intersection');
 
-function containsAllMatcherContext(splitValue /*: array */) /*: Function */ {
-  return function containsAllMatcher(value /*: array */) /*: boolean */ {
-
+function containsAllMatcherContext(ruleAttr /*: array */) /*: Function */ {
+  return function containsAllMatcher(runtimeAttr /*: array */) /*: boolean */ {
     let containsAll = false;
 
-    // If value has less elements than whitelist, there is now way that it contains all the whitelist elems.
-    if (value.length >= splitValue.length) {
-      containsAll = intersection(value, splitValue).length === splitValue.length;
+    // If runtimeAttr has less elements than whitelist, there is now way that it contains all the whitelist elems.
+    if (runtimeAttr.length >= ruleAttr.length) {
+      containsAll = intersection(runtimeAttr, ruleAttr).length === ruleAttr.length;
     }
 
-    log(`[containsAllMatcher] ${value} contains all elements of ${splitValue}? ${containsAll}`);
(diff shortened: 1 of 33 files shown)


src/engine/matchers/cont_all.js
```diff
@@ -19,25 +19,18 @@ limitations under the License.
 
 const log = require('debug')('splitio-engine:matcher');
 const intersection = require('lodash/intersection');
-const isArray = require('lodash/isArray');
 
-function containsAllMatcherContext(vo /*: whitelistObject */) /*: Function */ {
+function containsAllMatcherContext(splitValue /*: array */) /*: Function */ {
```
dendril: Please use a better variable name `whitelist` or `matchList` are options

src/engine/matchers/cont_str.js
```diff
@@ -0,0 +1,33 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+**/
+'use strict';
+
+/*eslint-disable eqeqeq */
```
dendril: Remove unused rule.

src/engine/matchers/eq_set.js
```diff
@@ -18,25 +18,16 @@ limitations under the License.
 /*eslint-disable eqeqeq */
```
dendril: Remove unused rule.

src/engine/parser/index.js
```diff
@@ -47,9 +47,8 @@ function parse(conditions: Array<Condition>, storage: SplitStorage): any {
       const matcher = matcherFactory(matcherDto, storage);
 
       return (key, attributes) => {
-        return matcher(
-          value(key, matcherDto.attribute, attributes) // value to be matched
-        );
+        const value = sanitizeValue(key, matcherDto, attributes);
```
dendril: ❤️

src/engine/transforms/matchers.js
```diff
@@ -41,54 +43,59 @@ function transform(matchers: Array<Matcher>): Array<ParsedMatcher> {
...
     } else if (type === matcherTypes.enum.WHITELIST) {
       value = whitelistTransform(whitelistObject);
     } else if (type === matcherTypes.enum.EQUAL_TO) {
-      value = unaryNumericObject;
+      value = numericTransform(unaryNumericObject);
 
       if (unaryNumericObject.dataType === 'DATETIME') {
-        unaryNumericObject.value = zeroSinceHH(unaryNumericObject.value);
+        value = zeroSinceHH(value);
+        inputProcessor = zeroSinceHH;
```
dendril: We should infer this just when needed.

Let discuss how we could move this into the `function sanitizeValue(...)`

src/engine/matchers/cont_str.js
NicoZelaya: Done

src/engine/matchers/cont_all.js
NicoZelaya: I've went for ruleAttr (the value set on the editor) and runtimeAttr (the value we receive from the dev).
We should get an agreement on a terminology for those two values, so it's easier to just read the code and understand which value is the one you're reading.

We can discuss wether it should be ruleAttr or ruleValue or ruleString/ruleSet/ruleNumber (and apply the same to runtimeAttr) because it would be an easy replace.

PR #73, opened 2017-05-02
[WIP] Time Tracker logger

NicoZelaya:
Time tracker logs could be either blue or magenta.

captura de pantalla 2017-05-05 a la s 12 46 57


patricioe:
@NicoZelaya thanks for this! Few comments:

1) Aside from ready took xx. Do we have splits was ready in XX and mysegments was ready in xx? In other words, the breakdown.
2) can we print how many impressions we sent and how long it took to post? Currently we just say Posting impressions which isn't very informative.
3) what do we log when a getTreatment is evaluated? Do we log the matchers used and what rule applied?

NicoZelaya:
@patricioe Look at below screenshot (extended output screenshot is on the top) and let me know if we need some other adjustments.

captura de pantalla 2017-05-05 a la s 12 48 38

patricioe:
@NicoZelaya awesome job!

PR #75, opened 2017-05-03
services/transport: use `status` rather than `statusText`

stephenmathieson:
In HTTP/2, the "reason" (`statusText`) is gone.

> 4.2 '"101" or "101 Switching Protocols"' - AIUI in HTTP/2 the reason
text is gone. The status pseudo header is numeric only.

ref:
http://lists.w3.org/Archives/Public/ietf-http-wg/2014JulSep/2461.html


patricioe:
Thank you for sending this along. We just published a 9.1.1 version with a small change to this approach to contemplate the 2XX family of responses.

PR #83, opened 2017-06-08
Adding support to send Full Story URL if available

HJewkes:
Before launching this must be wrapped in a configuration option

### DIFF
--- a/src/services/impressions/dto.js
+++ b/src/services/impressions/dto.js
@@ -36,6 +36,10 @@ module.exports = {
           if (entry.label) keyImpression.label = entry.label;
           if (entry.bucketingKey) keyImpression.bucketingKey = entry.bucketingKey;
 
+          properties = {};
+          if (FS) properties.full_story_url = FS.getCurrentSessionURL();
+          keyImpression.properties = JSON.stringify(properties);
+
           return keyImpression;
         })
       });


src/services/impressions/dto.js
```diff
@@ -36,6 +36,10 @@ module.exports = {
           if (entry.label) keyImpression.label = entry.label;
           if (entry.bucketingKey) keyImpression.bucketingKey = entry.bucketingKey;
 
+          properties = {};
+          if (FS) properties.full_story_url = FS.getCurrentSessionURL();
```
HJewkes: Not sure if I also need to check `if(typeof FS !== 'undefined)`

HJewkes: Ideally we would add a check against the configuration if full story integration is enabled

src/services/impressions/dto.js
```diff
@@ -36,6 +36,10 @@ module.exports = {
           if (entry.label) keyImpression.label = entry.label;
           if (entry.bucketingKey) keyImpression.bucketingKey = entry.bucketingKey;
 
+          properties = {};
```
NicoZelaya: Please add the `const` keyword. In JavaScript, declaring a variable without any keyword, would result in a global variable.
You could use const as this properties variable will always contain that object.

src/services/impressions/dto.js
NicoZelaya: Running the current check I think it's enough. You are checking for a _truthy_ value. If it's falsey (undefined, false, null...) so it may be better.

Please state on that file that you are using a global variable called FS with the following comment on top of the file (after the license disclosure):
` /* global FS */ `
With that we avoid linters complaining about FS not being defined.

src/services/impressions/dto.js
NicoZelaya: Please add the `const` keyword. In JavaScript, declaring a variable without any keyword, would result in a global variable.
You could use const as this properties variable will always contain that object.

PR #87, opened 2017-07-03
Boolean / Regex Matchers

dendril:
### Boolean matchers
- [ ] Define how we are going to evaluates none boolean attributes.
- [ ] Define how we are going to evaluates none boolean attributes coming from the server.

### Regex matchers
- [ ] What we are going to do when the attribute is not a string?
- [ ] How we are going to handle characters escaping?
- [ ] Unicode https://mathiasbynens.be/notes/es6-unicode-regex
- [ ] We don't have a way to inject flags so we are forcing to match cases right in the regex expression


src/engine/matchers/string.js
```diff
@@ -0,0 +1,40 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+**/
+'use strict';
+
+const log = require('../../utils/logger')('splitio-engine:matcher');
+
+function stringMatcherContext(ruleAttr /*: regex */) /*: Function */ {
+  return function stringMatcher(runtimeAttr /*: string */) /*: bool */ {
```
NicoZelaya: shouldn't this be matchesStringMatcher ? Naming it only string seems like something that may be confusing down the line. And also it's not respecting the SDK Spec name. 

If we could update the name (with the necessary updates on the file name, require statements, etc) I think it would be easier to read down the line.

src/engine/matchers/string.js
```diff
@@ -0,0 +1,40 @@
...
+
+    try {
+      re = new RegExp(ruleAttr);
+    } catch(e) {
+      log.debug(`[stringMatcher] ${ruleAttr} is an invalid regex`);
+
+      return false;
+    }
+
+    let regexMatches = re.test(runtimeAttr);
```
NicoZelaya: use const

dendril:
@NicoZelaya closing because I already merge everything into development

PR #91, opened 2017-07-27
Improve metrics

NicoZelaya:
Details of Metrics for the different parts of the system.

* latency => sdk.ready
* latency => sdk.getTreatment

* latency => splitChangeFetcher.time
* count   => splitChangeFetcher.status.XXX
* count   => splitChangeFetcher.exception

* latency => segmentChangeFetcher.time
* count   => segmentChangeFetcher.status.XXX
* count   => segmentChangeFetcher.exception

* latency => mySegmentsFetcher.time
* count   => mySegmentsFetcher.status.XXX
* count   => mySegmentsFetcher.exception


src/index.js
```diff
@@ -63,6 +63,8 @@ function SplitFactory(settings: Settings, storage: SplitStorage, gateFactory: an
   // Start background jobs tasks
   producer && producer.start();
   metrics && metrics.start();
+  // Add a callback to the ready latency tracker.
+  metrics && tracker.setupTrackers(metrics.trackers);
```
dendril: @NicoZelaya lets talk about this call

src/metrics/Logs.js
```diff
@@ -0,0 +1,69 @@
+class MetricsCollector {
```
dendril: Maybe we can rename the file or the class to align naming.

src/metrics/index.js
```diff
@@ -19,33 +19,56 @@ limitations under the License.
...
+      metricsCountersServiceRequest(settings, {
+        body: JSON.stringify(metricsDTO.fromCountersCollector(storage.count))
+      }
+    ))
+    .then(() => storage.count.clear())
+    .catch(() => storage.count.clear());
+
+    return Promise.all([latenciesPromise, countersPromise]).then(resp => {
+      // After both finishes, track the end and return the results
+      tracker.stop(tracker.C.METRICS_PUSH);
```
dendril: Lets review this with @sarrubia because I'm not 100% sure we should be tracking in this way.

src/metrics/index.js
```diff
@@ -87,6 +110,14 @@ const MetricsFactory = (settings: Object, storage: SplitStorage): Startable => {
...
       stopImpressionsPublisher && stopImpressionsPublisher();
       stopPerformancePublisher && stopPerformancePublisher();
+    },
+
+    // Metrics trackers
+    trackers: {
+      segmentChangesMetrics: new SegmentChangesMetrics(storage),
+      splitChangesMetrics: new SplitChangesMetrics(storage),
+      mySegmentsMetrics: new MySegmentsMetrics(storage),
+      sdkMetrics: new SDKMetrics(storage)
```
dendril: Maybe we can simplify names here

src/storage/Keys.js
```diff
@@ -1,80 +1,91 @@
...
+    return `${this.buildVersionablePrefix()}/impressions`;
   }
 
-  buildImpressionsKey(splitName: string) {
+  buildImpressionsKey(splitName) {
     return this.buildImpressionsKeyPrefix() + `.${splitName}`;
   }
 
-  buildLatencyKeyPrefix(): string {
+  buildLatencyKeyPrefix() {
```
dendril: Call `this.buildVersionablePrefix()`

src/storage/browser.js
```diff
@@ -32,6 +34,8 @@ const BrowserStorageFactory = (settings) => {
             segments: new SegmentCacheInMemory(childKeyBuilder),
             impressions: this.impressions,
             metrics: this.metrics,
+            // @TODO review this because I'm not sure this will work with shared instances
+            count: this.count,
```
dendril: TODO

src/storage/browser.js
```diff
@@ -67,6 +72,8 @@ const BrowserStorageFactory = (settings) => {
             segments: new SegmentCacheInLocalStorage(childKeysBuilder),
             impressions: this.impressions,
             metrics: this.metrics,
+            // @TODO review this because I'm not sure this will work with shared instances
+            count: this.count,
```
dendril: TODO

src/utils/timeTracker/index.js
```diff
@@ -0,0 +1,188 @@
...
+const thenable = require('../promise/thenable');
+
+// logger to be used on this module
+const logger = Logger.create('[TIME TRACKER]', {
+  showTimestamp: false,
+  showLevel: false,
+  useColors: false
+});
+// Map we will use for storing timers
+const timers = {};
```
dendril: Could we have collisions having multiple SDK instances?

src/utils/timeTracker/index.js
```diff
@@ -0,0 +1,188 @@
...
+    // as stopping the task when the promise resolves. Then return the promise
+    if (thenable(promise)) {
+      result = this.__attachToPromise(promise, task, taskUniqueId);
+    } else {
+      // If not, we return the stop function, as it will be stopped manually.
+      result = this.stop.bind(this, task, taskUniqueId);
+    }
+
+    // We start the timer, with an uniqueId attached to it's name.
+    timers[task + taskUniqueId] = timer();
```
dendril: I guess seeing this line I'm answering myself about the multiple instance question before.

src/utils/timeTracker/index.js
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param {string} task - The task we are starting.
+   * @param {string} modifier - (optional) The modifier for that specific task.
+   */
+  stop(task, modifier) {
+    const timerName = typeof modifier === 'string' ? task + modifier : task;
+    const timer = timers[timerName];
+    if (timer) {
+      // Stop the timer and round result for readability.
+      const et = Math.round( timer() );
```
dendril: Does it make sense to have `Math.round` here instead of inside the timer?

src/utils/timeTracker/index.js
```diff
@@ -0,0 +1,188 @@
...
+    metricTrackers = trackers;
+  },
+  /**
+   * Stops the tracking of a given task.
+   *
+   * @param {string} task - The task we are starting.
+   * @param {string} modifier - (optional) The modifier for that specific task.
+   */
+  stop(task, modifier) {
+    const timerName = typeof modifier === 'string' ? task + modifier : task;
```
dendril: Key generation should be moved into a function or method.

src/utils/timeTracker/index.js
```diff
@@ -0,0 +1,188 @@
...
+      if (tracker) {
+        // If we have a callback, we call it with the elapsed time of the task and then delete the reference.
+        tracker[tracker.method](et);
+      }
+    }
+  },
+  /**
+   * The constants shortcut for the task names.
+   */
+  C: CONSTANTS
```
dendril: Try a better name for constants. Maybe `LoggerNames` or `MetricNames` could work.

src/index.js
NicoZelaya: Already talked about it.

src/utils/timeTracker/index.js
NicoZelaya: Yes, as we've spoken before. That's still pending to be solved, but it's not a new problem. I'll try to get a solution before merging this branch either way.

src/storage/browser.js
NicoZelaya: Can you give me some more context on why you added these TODO? I don't remember why you've left them.

src/utils/timeTracker/index.js
NicoZelaya: I guess that I can move it inside the timer. It was outside because the `timer` utility was meant to be generic. But if we reuse it somewhere else (since this changes the only one using it is the timeTracker) I guess that the decimals won't be important, and we can always change.

I'll move it to the timer.

src/utils/timeTracker/index.js
NicoZelaya: Remember that the metrics are not the only usage. Maybe `TaskNames` ?

src/metrics/index.js
NicoZelaya: We'll need to follow up on this because I'm not following. This is one of the times we don't track as metric, but we output to the console. 

captura de pantalla 2017-08-02 a la s 17 27 01

PR #103, opened 2017-08-23
Context "capsule"

NicoZelaya:
Implementing a module for concentrating all modules/utilities we need to have on an SDK instance context, without conflicts between them.

The idea is to reduce the parameters overhead and just send the context to any place we may need something instance specific.

The next step would be to work towards removing code overhead caused by us trying to store things in a context, but this will be done at a later time.

### DIFF
--- a/src/utils/context/index.js
+++ b/src/utils/context/index.js
@@ -20,6 +20,27 @@ class Context {
     this.map = {};
   }
   /**
+   * Gets an item in the context instance or a promise if the item is not yet stored.
+   * @param {string} name - The name of the item we want to get
+   */
+  get(name) {
+    if (typeof name !== 'string' || typeof name === 'string' && !name.length) {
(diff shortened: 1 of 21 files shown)


src/utils/context/index.js
```diff
@@ -0,0 +1,77 @@
...
+    // Item already exists and no one is waiting for the value. Abort and return false.
+    if (existingItem !== undefined && typeof existingItem.manualResolve !== 'function') return false;
+
+    // Someone is waiting for this item, resolve to it.
+    if (thenable(existingItem) && existingItem.manualResolve) {
+      existingItem.manualResolve(item);
+    }
+
+    // We are storing a promise, when resolving save the value.
+    if (thenable(item)) item.then((item) => {
```
marciocaraballo: No need to use`( )` if we have only one param

marciocaraballo:
LGTM

PR #114, opened 2017-11-27
Fix CDN urls

tiago:


NicoZelaya:
Thank you for your contribution @tiago ! I've just updated the versions as well. I'll proceed with the merge.

PR #115, opened 2017-12-04
Update/readme dec 4 17

awgr:


awgr:
Will see what went sideways with the rebase

PR #118, opened 2017-12-04
[update] README: GO SDK + boilerplate

awgr:


NicoZelaya:
@awgr Codeship failure is there because there's a command added to the testing flow there that's still not merged into master `package.json`.

You can go ahead and merge, but if you want to make the build pass we can remove that test command from Codeship steps and just run it again. Just give me a notice so I can add that back later.

PR #121, opened 2017-12-14
[generated] update SDK labels

awgr:


NicoZelaya:
@awgr let me know when are we updating this to the QoS's

PR #130, opened 2018-02-14
SPLIT-3347 - migration of isomorphic-fetch to axios

marquezmanu:
[SPLIT-3347](https://splitio.atlassian.net/browse/SPLIT-3347)

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,3 +1,28 @@
+## 10.1.0 (March XX, 2018)
+
+* Replaced isomorphic-fetch with axios.
+
+  Isomorphic-fetch was deprecated around end of 2015, and many dependencies have
+  updated, so and old lib with updated dependencies is causing trouble.
+  Libraries evolve all the time. This library in particular was maintained until
+  end of 2015. Two years later the JS ecosystem changed a lot, so the library
+  itself has old code. In fact, libraries that depend on these one have been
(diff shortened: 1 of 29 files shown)


NEWS.txt
```diff
@@ -1,3 +1,28 @@
+## 10.1.0 (March XX, 2018)
+
+* Replaced isomorphic-fetch with axios.
+
+  Isomorphic-fetch was deprecated around end of 2015, and many dependencies have
```
marciocaraballo: I'd remove mentioning that is was deprecated around 2015. Maybe we can just explain that we changed it because it was deprecated, but not so long ago ? I'm not sure how specific has this explanation to be

NicoZelaya:
Don't forget to check that all headers and request settings are the same than before the change.

src/utils/settings/index.js
```diff
@@ -127,7 +132,7 @@ const proto = {
   /**
    * Returns a settings clone with the key and traffic type (if provided) overriden.
    * @param {SplitKey} key
-   * @param {string} [trafficType]
```
NicoZelaya: Just FYI the `[paramName]` is a way to reflect that param is nullable/optional. 
Not too important internally but it is for externally facing functions, mostly for TS declarations, as the comment will appear for intellisense.

PR #133, opened 2018-03-12
DEV-4481 - redis return control and exception label on error

marquezmanu:
- return control and label exception in case that redis has a connection error


src/__tests__/node_redis.spec.js
```diff
@@ -12,45 +16,113 @@ const config = {
...
 
-  assert.end();
+      client.destroy();
 
+      // close server connection
+      server.close().then(assert.end);
+    });
 });
+
+tape('NodeJS Redis / Connection Error', async function (assert) {
```
NicoZelaya: Add a test for the impression label === 'exception'

PR #150, opened 2018-04-27
Input validation

marquezmanu:
NOT MERGE UNTIL ALL TESTING APPS ARE GREEN

- input validations on getTreatments, track, manager methods
- udpated node specific unit test that were outdated for a long time ago
- fixed memory leak on events factory, a setTimeout was never removed before a stop method was called.
- not push impressions if the matching key is invalid
- improved key parser, now it will log if the key is invalid, will try to sanatize the key if possible and return false instead to throw a js exception.


src/__tests__/destroy/browser.spec.js
```diff
@@ -66,25 +66,21 @@ tape('SDK destroy for BrowserJS', async function (assert) {
     const events = JSON.parse(request.data);
 
     /* 4 events were pushed */
```
NicoZelaya: keep comments updated for future reference

src/__tests__/destroy/browser.spec.js
```diff
@@ -45,7 +45,7 @@ tape('SDK destroy for BrowserJS', async function (assert) {
   const manager = factory.manager();
 
   // Events are shared between shared instances.
-  assert.true(client.track('tt', 'eventType' /* Invalid value is stored as 0 */));
+  assert.ok(client.track('tt', 'eventType', 'invalid value' /* Invalid values are not tracked */) === false);
```
NicoZelaya: you could use equal

src/__tests__/node.spec.js
```diff
@@ -0,0 +1,52 @@
...
+mock.onGet(settings.url('/mySegments/marcio@split.io')).reply(200, mySegmentsMarcio);  
+
+tape('## Node JS - E2E CI Tests ##', async function (assert) {
+  /* Check client evaluations. */  
+  assert.test('E2E / In Memory', evaluationsSuite.bind(null, config, key));
+
+  /* Check impressions */
+  assert.test('E2E / Impressions', impressionsSuite.bind(null, key, mock));
+
+  // TODO: add events and shared instantiation e2e tests
```
NicoZelaya: Question: What's the blocker for adding them now?

src/utils/__tests__/track/index.spec.js
```diff
@@ -0,0 +1,108 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+**/
+import tape from 'tape';
+import validate from '../../track/validate';
+
+tape('TRACK VALIDATE / if valid arguments are passed should return true', assert => {
+  const key = 'some_key';
```
NicoZelaya: Just in the sake of not having more and more tests lines (we already have a ton) there's no need for creating a variable if you're not going to reuse it on an assert or something. 

Also you can kind of group the test cases for valid and invalid in two different test context, and vary the message that says: `track validate will return true if all params are valids`.

For example:
tape('TRACK VALIDATE / If valid arguments are passed it should always return true, assert => {
   assert.ok(areValidTrackArguments(...args1), 'Should be valid if everything is defined with valid types');
   assert.ok(areValidTrackArguments(...args2), 'Should be valid if everything is defined but the value is undefined');
  assert.ok(areValidTrackArguments(...args3), 'Should be valid if everything is defined but the value is null');
 ....
})

src/utils/key/factory.js
```diff
@@ -1,25 +1,39 @@
...
-    }
-
     if (isObject(key)) {
-      if (!key[keyProperty]) {
-        throw `key should has property ${keyProperty}`;
-      }
+      const sanatizedProperty = sanatize(key[keyProperty]);
 
-      return key[keyProperty];
+      if (sanatizedProperty !== false) {
```
NicoZelaya: Unnecessary condition. If it's false we return false, if it's not false we return the value. Just return the `sanitizedProperty`

src/utils/manager/validate.js
```diff
@@ -0,0 +1,34 @@
...
+const log = logFactory('splitio-client');
+
+function managerValidate(splitName) {
+  if (splitName === null || splitName === undefined) {
+    log.error('split: split_name cannot be null');
+    return false;
+  }
+
+  if (!isString(splitName)) {
+    log.error('split: split_name X must be a string');
```
NicoZelaya: Shouldn't be `manager: ...` instead of `split: ...` ? Since it will be more helpful for live customer debugging (which is the main point of the logger functionality)

src/__tests__/node.spec.js
marquezmanu: we could but it took me a lot of time to update this tests

PR #158, opened 2018-05-30
10.3.1 - Adding events as a dependency.

NicoZelaya:
There are some cases where we can't use the `events` module from NodeJS, like on the Browser or transpiling for mobile with React Native. 
For the `AMD` build we're taking care, but for `cjs` or `esm` builds the customer may need to install the package. Adding it as a dependency simplifies that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@splitsoftware/splitio",
-  "version": "10.3.0",
+  "version": "10.3.1-canary.0",
   "description": "Split SDK",
   "files": [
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -94,6 +94,7 @@
     "@types/node": "9.4.7",
     "axios": "0.18.0",
     "babel-runtime": "6.26.0",
+    "events": "^3.0.0",
```
marquezmanu: set fixed version 3.0.0

PR #175, opened 2018-08-31
chore(package): update babel-loader to version 8.0.0

NicoZelaya:
Closes #156


NicoZelaya:
Not removing support for old node version ATM.

PR #193, opened 2018-09-18
[SDKs-140] Flush on SIGTERM for NodeJS

NicoZelaya:
The idea was to add automatic support for graceful shutdown. Started with Node.


mmelograno:
LGTM!

PR #266, opened 2019-05-14
 Fix import/export syntax for labels

### LINKED ISSUE #265: Incorrect ES6 import syntax causing warnings and missing data
This file is defined/imported elsewhere incorrectly: 
https://github.com/splitio/javascript-client/blob/development/src/utils/labels/index.js

It causes issues for at least the following lines.
https://github.com/splitio/javascript-client/blob/development/src/client/client.js#L11
https://github.com/splitio/javascript-client/blob/development/src/utils/inputValidation/split.js#L2

The files with issues are trying to use destructuring assignment on an import statement, but that's not a valid syntax. Brackets are only to be used for named exports.

kellym:
Fixes #265.

### DIFF
--- a/src/engine/__tests__/evaluator/index.spec.js
+++ b/src/engine/__tests__/evaluator/index.spec.js
@@ -15,7 +15,7 @@ limitations under the License.
 **/
 import tape from 'tape';
 import evaluator from '../../evaluator';
-import LabelsConstants from '../../../utils/labels';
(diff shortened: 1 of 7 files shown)


NicoZelaya:
Thank you very much for working on this PR! I'll go ahead and move this version to our automation environment and releasing a patch soon. 

I'll let you know on the issue once it's out!

kellym:
Great, thanks!

PR #283, opened 2019-07-16
[SDKS-836]: TrafficType on updated split

mmelograno:
# Javascript Client

## Tickets covered:
* [SDKS-836: TrafficType on updated split](https://splitio.atlassian.net/browse/SDKS-836)

## What did you accomplish?
* Added logic to decrement tt if split exists previously to be added on updates.
* Added `JSON.parse` when a split is received from addSplit and from cache to detect the tt and decrement/increment it in the tt cache storage.

## How to test new changes?
* Run command `npm test `

## Version Release
* `10.8.2`

### DIFF
--- a/src/storage/SplitCache/InLocalStorage.js
+++ b/src/storage/SplitCache/InLocalStorage.js
@@ -1,6 +1,7 @@
-import { isFinite } from '../../utils/lang';
+import { isFinite, toNumber } from '../../utils/lang';
 import logFactory from '../../utils/logger';
 const log = logFactory('splitio-storage:localstorage');
+
 
 class SplitCacheLocalStorage {
 
@@ -12,13 +13,12 @@ class SplitCacheLocalStorage {
     try {
       if (split && split.trafficTypeName) {
         const ttKey = this.keys.buildTrafficTypeKey(split.trafficTypeName);
-        const count = Number(localStorage.getItem(ttKey)) - 1;
+        const count = toNumber(localStorage.getItem(ttKey)) - 1;
         if (count > 0) localStorage.setItem(ttKey, count);
         else localStorage.removeItem(ttKey);
       }
     } catch (e) {
       log.error(e);
-      return;
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/storage/SplitCache/InLocalStorage.js
```diff
@@ -129,7 +146,7 @@ class SplitCacheLocalStorage {
   }
 
   trafficTypeExists(trafficType) {
-    const ttCount = localStorage.getItem(this.keys.buildTrafficTypeKey(trafficType));
+    const ttCount = Number(localStorage.getItem(this.keys.buildTrafficTypeKey(trafficType)));
```
NicoZelaya: FYI you might have used the toNumber utility we have. (Number constructor works well here so just FYI).

src/storage/SplitCache/InLocalStorage.js
```diff
@@ -8,13 +8,32 @@ class SplitCacheLocalStorage {
...
+  decrementTrafficType(split) {
+    try {
+      if (split && split.trafficTypeName) {
+        const ttKey = this.keys.buildTrafficTypeKey(split.trafficTypeName);
+        localStorage.setItem(ttKey, Number(localStorage.getItem(ttKey)) - 1);
+        if (!Number(localStorage.getItem(ttKey))) delete localStorage[ttKey];
+      }
+    } catch (e) {
+      log.error(e);
+      return;
```
NicoZelaya: Unnecessary return

src/storage/SplitCache/InLocalStorage.js
```diff
@@ -8,13 +8,32 @@ class SplitCacheLocalStorage {
     this.keys = keys;
   }
 
+  decrementTrafficType(split) {
+    try {
+      if (split && split.trafficTypeName) {
+        const ttKey = this.keys.buildTrafficTypeKey(split.trafficTypeName);
+        localStorage.setItem(ttKey, Number(localStorage.getItem(ttKey)) - 1);
+        if (!Number(localStorage.getItem(ttKey))) delete localStorage[ttKey];
```
NicoZelaya: This lines could be something around:
```
const newCount = Number(localStorage.getItem(ttKey) - 1;
if (newCount > 0) localStorage.setItem(ttKey, newCount)
else localStorage.removeItem(ttKey)
```

Ideally we avoid the double operation with localstorage, also it's more correct to use the removeItem function from it instead of just cleaning the reference.

PR #307, opened 2019-09-27
Setting IPAddressesEnabled to disable sending IP and machine name to BE. 

EmilianoSanchez:
- Setting IPAddressesEnabled only for Nodejs. True by default.
- Setting value used as condition for sending headers 'SplitSDKMachineIP' with IP address, and 'SplitSDKMachineName' with machine name.

### DIFF
--- a/src/__tests__/nodeSuites/ip-addresses-setting.spec.js
+++ b/src/__tests__/nodeSuites/ip-addresses-setting.spec.js
@@ -1,11 +1,13 @@
 import { SplitFactory } from '../../';
 import SettingsFactory from '../../utils/settings';
 import runtime from '../../utils/settings/runtime';
+import splitChangesMock1 from '../mocks/splitchanges.since.-1.json';
 
 // Header keys and expected values. Expected values are obtained with the runtime function evaluated with IPAddressesEnabled in true.
 const HEADER_SPLITSDKMACHINEIP = 'SplitSDKMachineIP';
 const HEADER_SPLITSDKMACHINENAME = 'SplitSDKMachineName';
-const { ip: HEADER_SPLITSDKMACHINEIP_VALUE, hostname: HEADER_SPLITSDKMACHINENAME_VALUE} = runtime(true);
+const { ip: HEADER_SPLITSDKMACHINEIP_VALUE, hostname: HEADER_SPLITSDKMACHINENAME_VALUE } = runtime(true);
+const NA = 'NA';
 
 // Refresh rates are set to 1 second to finish the test quickly. Otherwise, it would finish in 1 minute (60 seconds is the default value)
 const scheduler = {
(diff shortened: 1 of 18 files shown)


types/splitio.d.ts
```diff
@@ -38,7 +38,8 @@ interface ISettings {
     authorizationKey: string,
     key: SplitIO.SplitKey,
     trafficType: string,
-    labelsEnabled: boolean
+    labelsEnabled: boolean,
+    IPAddressesEnabled: boolean
```
NicoZelaya: Let's add a test for it too.

src/__tests__/nodeSuites/ip-addresses-setting.spec.js
```diff
@@ -52,44 +54,52 @@ const configWithIPAddressesDefault = {
...
-  });
+  function mockAndAssertIPAddressesEnabled(config, endpointToMock, clientAction){
+    finish.next();
 
-  clientDisabled.track('nicolas@split.io', 'sometraffictype', 'someEvent', 10);
+    const splitio = SplitFactory(config);
+    const client = splitio.client();
+    const settings = SettingsFactory(config);
+  
+    mock.onPost(settings.url(endpointToMock)).replyOnce(req => {
```
NicoZelaya: We should try to mock all endpoints where we post data, for events, impressions and both for telemetry (counts and times).

You can either use a generator function or keep track of these with a counter of some sorts, as long as you do it inside the `mockAndAssertIPAddressesEnabled` function.

src/__tests__/nodeSuites/ip-addresses-setting.spec.js
```diff
@@ -52,44 +54,52 @@ const configWithIPAddressesDefault = {
...
+    }while(count);
+    clients.forEach(client => client.destroy());
+    assert.end();
+  })();
 
-  let splitioDisabled = SplitFactory(configWithIPAddressesDisabled);
-  let clientDisabled = splitioDisabled.client();
-  const settingsDisabled = SettingsFactory(configWithIPAddressesDisabled);
+  function assertHeaders(IPAddressesEnabled, req){
+    assert.equal(HEADER_SPLITSDKMACHINEIP in req.headers, IPAddressesEnabled, `Request must include ${HEADER_SPLITSDKMACHINEIP} header if IPAddressesEnabled is true.`);
```
NicoZelaya: Truly a minor thing but the assertion message is missleading (it always talks about the value being true). Maybe we can have something like:
`Request must ${IPAddressesEnabled ? '' : 'NOT '}include ${HEADER_SPLITSDKMACHINEIP} header if IPAddressesEnabled is ${IPAddressesEnabled}.`

NicoZelaya:
Let's add a similar integration suite for Browsers to validate the IP is not sent regardless of what's on the config object.

src/__tests__/nodeSuites/ip-addresses-setting.spec.js
EmilianoSanchez: Done

src/__tests__/nodeSuites/ip-addresses-setting.spec.js
EmilianoSanchez: Done

NicoZelaya:
LGTM, @EmilianoSanchez go ahead and merge whenever you want

EmilianoSanchez:
* Fixed typescript definition (Removed 'LOCALSTORAGE' value from NodeSyncStorage type).
* Include unit and integration tests.

PR #309, opened 2019-10-10
Implementation and E2E tests of BrowserListener with Beacon API

EmilianoSanchez:
# JS SDK

## What did you accomplish?
BrowserListener registers a handler for 'unload' DOM events, in order to flush remaining impressions and events when a page is closed or reload, using Beacon API if available or falling back to Axios (XHR).

Three E2E tests were added:
1- Check that sendBeacon is not called when no events are impressions are in cache
2- Check that payloads of beacon requests are well formed
2- Check that HTTP requests are sent using Axios when Beacon API is not available, and also check if their payloads are well formed.

## Extra Notes
Manual testing (opening and closing an HTML page in a browser to trigger an 'unload' event) works well with the new beacon API feature.
Fallback to Axios works well on IE10

### DIFF
--- a/src/listeners/__tests__/browser.spec.js
+++ b/src/listeners/__tests__/browser.spec.js
@@ -1,0 +1,46 @@
+import tape from 'tape-catch';
+import sinon from 'sinon';
+import Context from '../../utils/context';
+import StorageFactory from '../../storage';
+import SettingsFactory from '../../utils/settings';
+import BrowserSignalListener from '../browser';
+
+const UNLOAD_DOM_EVENT = 'unload';
+
+const windowAddEventListenerSpy = sinon.spy();
+const windowRemoveEventListenerSpy = sinon.spy();
+
+sinon.stub(window, 'addEventListener').callsFake(windowAddEventListenerSpy);
+sinon.stub(window, 'removeEventListener').callsFake(windowRemoveEventListenerSpy);
+
+const createFakeContext = function() {
+  const config = {};
+  const context = new Context();
+
+  const settings = SettingsFactory(config);
+  context.put(context.constants.SETTINGS, settings);
+
+  const storage = StorageFactory(context);
+  context.put(context.constants.STORAGE, storage);
+  return context;
+};
+
+tape('Browser JS / Browser listener class constructor, start and stop methods', function (assert) {
+  const listener = new BrowserSignalListener(createFakeContext());
+
+  listener.start();
+
+  // Assigned right function to right signal.
+  assert.ok(windowAddEventListenerSpy.calledOnce);
+  assert.ok(windowAddEventListenerSpy.calledOnceWithExactly(UNLOAD_DOM_EVENT, listener.flushData));
+
+  // pre-check and call stop
+  assert.ok(windowRemoveEventListenerSpy.notCalled);
+  listener.stop();
+
+  // removed correct listener from correct signal on stop.
+  assert.ok(windowRemoveEventListenerSpy.calledOnce);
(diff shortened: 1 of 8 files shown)


src/listeners/browser.js
```diff
@@ -1,16 +1,87 @@
...
+  /**
+   * start method. 
+   * Called when SplitFactory is initialized. 
+   * We add a handler on unload events. The handler flushes remaining impressions and events to the backend.
+   */ 
   start() {
-    // noop
+    log.debug('Registering flush handler when unload page event is triggered.');
+    if (window && window.addEventListener) {
+      window.addEventListener('unload', this._flushEventsAndImpressions);
```
mmelograno: make `unload` as a parameter of the class `BrowserSignalListener` in case you subscribe to different event in the future

src/listeners/browser.js
```diff
@@ -1,16 +1,87 @@
...
+  }
+
+  /**
+   * _flushEventsAndImpressions method. 
+   * Called when unload event is triggered. It flushed remaining impressions and events to the backend, 
+   * using beacon API if possible, or falling back to XHR.
+   */ 
+  _flushEventsAndImpressions() {
+    // if there are impressions in storage, send them to backend
+    if (!this.storage.impressions.isEmpty()) {
```
mmelograno: wouldn't be good idea to split flushEvent and flushImpressions and rename `_flushEventsAndImpressions` to `_fliushData` and call `flushEvent` and `flushImpressions` inside that funtion?

src/listeners/__tests__/browser.spec.js
```diff
@@ -0,0 +1,46 @@
...
+import StorageFactory from '../../storage';
+import SettingsFactory from '../../utils/settings';
+import BrowserSignalListener from '../browser';
+
+const UNLOAD_DOM_EVENT = 'unload';
+
+const windowAddEventListenerSpy = sinon.spy();
+const windowRemoveEventListenerSpy = sinon.spy();
+
+sinon.stub(window, 'addEventListener').callsFake(windowAddEventListenerSpy);
```
NicoZelaya: You could have only used a spy (or a stub if you have to prevent the real method being executed)

src/listeners/__tests__/browser.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const settings = SettingsFactory(config);
+  context.put(context.constants.SETTINGS, settings);
+
+  const storage = StorageFactory(context);
+  context.put(context.constants.STORAGE, storage);
+  return context;
+};
+
+tape('Browser JS / Browser listener class constructor, start and stop methods', function (assert) {
```
NicoZelaya: Let's test the rest of the methods too.

src/listeners/browser.js
```diff
@@ -1,16 +1,101 @@
...
   }
 
+  /**
+   * start method. 
+   * Called when SplitFactory is initialized. 
+   * We add a handler on unload events. The handler flushes remaining impressions and events to the backend.
+   */
   start() {
-    // noop
+    log.debug('Registering flush handler when unload page event is triggered.');
```
NicoZelaya: Move the log inside the if, since it could be logged but not be true. That would be very confusing when we're debugging with a customer.

Consider adding a message with an else clause too

src/listeners/browser.js
```diff
@@ -1,16 +1,101 @@
...
 
+
+  /**
+   * stop method. 
+   * Called when client is destroyed. 
+   * We need to remove the handler for unload events, since it can break if called when Split context was destroyed.
+   */ 
   stop() {
-    // noop
+    log.debug('Deregistering flush handler when unload page event is triggered.');
```
NicoZelaya: Move the log here too. Consider adding a message with an else clause too.

PR #310, opened 2019-10-11
Feature/mget treatments

mmelograno:
# JS SDK

## What did you accomplish?
= Added `mget` for multiple evaluations at once.
- Added new method `evaluateFeatures` for doing multiple evaluations at once.
- Refactored `evaluator` for evaluating one feature and multiple features.

- Check redis calls - it should execute one call instead of one per each feature.

## Extra Notes


src/storage/SplitCache/InLocalStorage.js
```diff
@@ -197,6 +197,18 @@ class SplitCacheLocalStorage {
     log.info('Flushing localStorage');
     localStorage.clear();
   }
+
+  /**
+   * Fetches multiple splits definitions.
+   */
+  fetchMany(splitNames) {
+    const splits = new Map();
+    for (let i = 0; i < splitNames.length; i++) {
```
EmilianoSanchez: Is there any optimization if using "for (let splitName in splitNames) {" instead of "for (let i = 0; i < splitNames.length; i++) {" ?

mmelograno: updated!

PR #315, opened 2019-10-23
Fix #314, Support @types/node@12.11.5

### LINKED ISSUE #314: Support for Nodejs 12
Hi, 
I tried running the package with node 12 in my front end application & it gave build errors. 

Please provide support. 

Thanks
Akanksha Gaur

codler:
Fix error message `Property 'on' does not exist on type 'IClient'.`

The latest node typings doesnt have NodeJS.Events anymore, it is now NodeJS.EventEmitter

### DIFF
--- a/types/splitio.d.ts
+++ b/types/splitio.d.ts
@@ -277,9 +277,9 @@ interface INodeBasicSettings extends ISharedSettings {
 /**
  * Common API for entities that expose status handlers.
  * @interface IStatusInterface
- * @extends NodeJS.Events
+ * @extends NodeJS.EventEmitter
  */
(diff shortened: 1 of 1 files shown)


codler:
@NicoZelaya Ah okej, i think EventEmitter exist in older node version so it should be fine

codler:
@NicoZelaya When will the next patch be release?

NicoZelaya:
Hi codler! It's going out between today and tomorrow. I've included this update on PR #311 which is now already tested and merged into development.
There's another small set of changes we'll include, test and then release. I'll let you know here when that happens if you'd like.

PR #326, opened 2019-11-18
return expire result to avoid unhandled promise rejection race condition

scamden:
# JS SDK

## Extra Notes

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,6 @@
+10.9.2 (Nov 19, 2019)
+ - Bugfixing - When using Redis, return expire result to avoid race condition with redis disconnection due to a client destroy.
+
 10.9.1 (Nov 7, 2019)
  - Bugfixing - Avoid certain keys (containing one of three keywords) on the browser to generate a request to the wrong endpoint.
  - Added SonarQube integration.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@splitsoftware/splitio",
-  "version": "10.9.1",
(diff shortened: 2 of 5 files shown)


NicoZelaya:
Hi @scamden ! Thank you very much for contributing with this PR. I'll take it from here and deploy to our stage environment while I add test coverage to prevent a regression in the future. 

Just to be on the same page than you, can you confirm the issue was due to the expire promise not being resolved before the track promise callback you attached runs? Or what behaviour did you fix with this return statement?

Thanks again and our apologies for the inconveniences. 
Nico.

scamden:
I believe what’s happening is that the split client gets destroyed which closes the redis connection and then because the promise can’t be / isn’t awaited here, the expire call causes a command to be sent to redis to expire after the connection had already closed (only sometimes, it’s a race). It causes a “Stream isn’t writable” error to be thrown from redis. That error can happen a few different ways unfortunately but this is the only one we could find your sdk and patching it in our version of your code fixed the intermittent error (which was occurring roughly 5-10% of the time on normal requests). Tough to write a test for since it’s intermittent and seems to be internal in the redis client but I’ll leave that part to you :)

NicoZelaya:
I've just added the test to your fork (sorry for the delay, today was busier than I expected). Would you mind taking a look @scamden ? 

Also @mmelograno would you mind giving this another review?

scamden:
lgtm :)

NicoZelaya:
Thanks both!

NicoZelaya:
@scamden version 10.9.2 is out with this fix. Thank you again for your contribution!!

scamden:
thanks!

Issue #350, opened 2020-01-22
Update axios version to 0.19.2


nadios:
Could you please update the `axios` dependency to the latest version (`0.19.2`) as version `0.19.1` has some issues (https://github.com/axios/axios/pull/2679, https://github.com/axios/axios/issues/2646) which we experience.

jpetersen-cuda:
+ 1

NicoZelaya:
Hi @nadios,  thank you for flagging!

We'll definitely take care of this on our upcoming release, will let you know here once it's out.

Best,
Nico.

Issue #337, opened 2020-01-03
SDK_UPDATE emitted in localhost mode when no split status changed


petriaarnio:
**Problem**
In localhost mode, SDK_UPDATE gets emitted in each refresh interval timeout even if there are no split changes. The functionality differs from the normal mode where SDK_UPDATE is emitted only for real changes.
This causes unwanted component re-rendering in React applications and is especially visible when rendering is a heavy operation e.g. map component.

**Suggestion**
Add split changes checking before emitting SDK_UPDATE.

Issue #335, opened 2020-01-01
How to silence console.errors from the SDK?


ethanneff:
**Problem**

`console.error()` interrupts React Native apps with a large overlay. This is problematic when debugging the app in offline mode. 

**Details**

I get the following error from the Split SDK when implementing a factory:

```
console.error: "[ERROR] Split SDK emitted SDK_READY_TIMED_OUT event."
```

```ts
const factory = SplitFactory({
  core: { authorizationKey, key }
});
```

![Screenshot_20191231-200804](https://user-images.githubusercontent.com/2933593/71637554-69d43a00-2c0b-11ea-8861-a9a23cab7d62.png)

**Question**

How do I silence `console.errors` from the Split SDK? Even when listening to the event, the `console.error` still comes through.

```ts
client.on(client.Event.SDK_READY_TIMED_OUT, function () {});
```

NicoZelaya:
Hi @ethanneff! 

I would like to start by offering you my most sincere apologies for dropping the ball on getting back to you, as well as ours for the inconveniences.

It's actually our plan to remove the calls to `console.error` and just stick with `console.log` method. We'll also treat these as just another log level instead of always printing.
While it's not an actual error or exception of any kind and it doesn't affect the execution flow (should only be a red message) we've recently learned that some client side tools might consider these as actual errors.

I'll make sure to speed up resolution for this one so we can test and make a patch release before end of week. 

Kindest regards,
Nicolas Zelaya.

ethanneff:
When was this fixed? Our app is currently using `10.10.0` and we are still getting the errors. #351 did not fix it. 

![Simulator Screen Shot - iPhone 11 - 2020-03-20 at 15 08 41](https://user-images.githubusercontent.com/2933593/77216078-10c2c680-6ad5-11ea-89f1-8b9a5e2fa65c.png)

NicoZelaya:
Hi @ethanneff ! Sorry I didn't want to tag you earlier to not generate a notification because I thought we reached out separately but yes, should be fixed. Can you try version `10.11.1` ? 

I'm curious why you say PR #351 didn't solve it, how are you testing that? Was included on the `10.11.0` release if I'm not mistaken.

In any case, give a try to `10.11.0` or `10.11.1` and you should be able to turn off error logs and besides, we're not using the `console.error` method anymore in order to fix these React Native issue.

I'll reopen until I hear back from you.
Thanks in advance!
Nico.

PR #351, opened 2020-01-23
Maintenance patch - #350, #335 and #337 plus dependency updates

NicoZelaya:
# JS SDK

## What did you accomplish?
- Update axios to version 0.19.2 in order to address #350 
- Updated logging logic related to #335
- Fixed the logic behind SDK_UPDATE events to only emit when there was an update addressing #337 
- Update some dependencies `@babel/runtime-corejs3` and `@types/node`
- Updated some dev dependencies `webpack`, `babel`, `sinon`, `eslint` and `copyfiles`.

The usual, run every test and build we have!

## Extra Notes
Aiming to release this week.


src/services/splitChanges/offline/browser.js
```diff
@@ -1,8 +1,32 @@
...
+  if (names.length !== Object.keys(previousMock).length) {
+    previousMock = currentData;
+    return true;
+  }
+
+  return names.some(name => {
+    const newSplit = !previousMock[name];
+    const newTreatment = previousMock[name] !== currentData[name];
+
+    previousMock = currentData;
```
EmilianoSanchez: outside the arrow function?

PR #356, opened 2020-02-10
Event Tracker and Integration Manager

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- `ImpressionTracker` and `ImpressionsTracker` (formerly `ImpressionsTrackerContext`) were renamed for consistency
- `EventTracker` was added to refactor client logic regarding event tracking
- `integrations` setting validation
- `IntegrationManager` was added to handle SDK integrations

- Unit tests for `EventTracker`
- Unit tests of `integrations` setting validation
- Unit tests for `IntegrationManager`

## Extra Notes

- Review: logging messages, type definitions


src/integrations/browser.js
```diff
@@ -0,0 +1,56 @@
...
+  const settings = context.get(context.constants.SETTINGS);
+
+  // If no integrations settings, not return a integration manager
+  if (!settings.integrations)
+    return;
+
+  const listeners = [];
+
+  for (const integrationOptions of settings.integrations) {
+    const { type } = integrationOptions;
```
NicoZelaya: These two lines become this for es5 build in npm (people that bundles) as well as our UMD bundle:

```
    for (var _iterator = _babel_runtime_corejs3_core_js_get_iterator__WEBPACK_IMPORTED_MODULE_0___default()(settings.integrations), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
      var integrationOptions = _step.value;
      var type = integrationOptions.type;
```

Why not use a regular for loop? Or a forEach?
```
for (let i=0; i < settings.integrations.length; i++) {
  const integrationOptions = settings.integrations[i];
  const type = integrationOptions.type;
}

settings.integrations.forEach(integrationOptions => {
   const type = integrationOptions.type;
  // ... function body
}
```

src/trackers/__tests__/impression.spec.js
```diff
@@ -17,29 +18,36 @@ const generateContextMocks = () => {
...
 
 class ContextMock {
-  constructor(fakeStorage, fakeSettings) {
+  constructor(fakeStorage, fakeSettings, fakeIntegrationsManager) {
     this.constants = {
-      STORAGE: 'storage',
-      SETTINGS: 'settings'
+      STORAGE,
+      SETTINGS,
+      INTEGRATIONS_MANAGER,
```
NicoZelaya: trailing comma.

src/utils/constants/index.js
```diff
@@ -15,4 +15,10 @@ export const CONTROL_WITH_CONFIG = {
 };
 // Constants for unknown and not-applicable values
 export const UNKNOWN = 'unknown';
-export const NA = 'NA';
\ No newline at end of file
+export const NA = 'NA';
+// Integration types
+export const GA_TO_SPLIT = 'GA_TO_SPLIT';
```
NicoZelaya: Question: Why not a constants file in integrations folder? because it's used in the settings too?

src/utils/settings/index.js
```diff
@@ -123,7 +127,11 @@ function defaults(custom) {
   setupLogger(withDefaults.debug);
 
   // Current ip/hostname information
-  withDefaults.runtime = runtime(withDefaults.core.IPAddressesEnabled,withDefaults.mode === CONSUMER_MODE);
+  withDefaults.runtime = runtime(withDefaults.core.IPAddressesEnabled, withDefaults.mode === CONSUMER_MODE);
+
+  // ensure a valid list of integrations.
+  // `integrations` returns `undefined` or a non-empty array of valid integration items.
+  withDefaults.integrations = integrations(withDefaults);
```
NicoZelaya: Why not always defaulting to have an array, but empty if the data was invalid, so after this point validations become easier?

src/utils/settings/integrations/__tests__/common.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      type: 'INT2',
+      param1: 'param1',
+      param2: 'param2',
+    };
+    const invalid = {
+      param3: 'param3',
+    };
+
+    assert.deepEqual(validateIntegrationsSettings(
+      { integrations: [valid, validWithOptions, invalid] }), undefined,
```
NicoZelaya: shouldn't we add some validations for incorrect types? like if you receive:

```
integrations: [false, 0, Infinity, new Error(), () => {}, []]
```

src/trackers/impression.js
```diff
@@ -18,38 +18,46 @@ import logFactory from '../utils/logger';
...
-        }
-      }, 0);
+      if (listener || integrationsManager) {
+        const impressionData = {
+          impression,
+          attributes,
+          ip,
+          hostname,
+          sdkLanguageVersion
+        };
```
NicoZelaya: Be careful. While the data is the same, every time an object is passed to a customer to consume we should aim to sending a copy, not the reference.
If we pass along the reference and something unexpected is modified on it, it might impact other pieces of the SDK.

src/client/client.js
```diff
@@ -134,13 +121,7 @@ function ClientFactory(context) {
     // This may be async but we only warn, we don't actually care if it is valid or not in terms of queueing the event.
     validateTrafficTypeExistance(trafficTypeName, context, 'track');
 
-    const tracked = storage.events.track(eventData, size);
-
-    if (thenable(tracked)) {
-      return tracked.then(queueEventsCallback.bind(null, eventData));
-    } else {
-      return queueEventsCallback(eventData, tracked);
-    }
```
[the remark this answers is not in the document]
EmilianoSanchez: is okey to keep this logic here? why not in EventTracker? for instance, ImpressionsTracker is responsable for tracking impressions into the storage and handling the response (async when using redis, and sync when memory or localstorage)

src/utils/settings/integrations/common.js
```diff
@@ -0,0 +1,25 @@
...
+const validateIntegrationsSettings = (settings, validIntegrationTypes = []) => {
+  const { integrations } = settings;
+
+  // If integrations is not an array or an empty array, we return undefined (no integrations).
+  if (!Array.isArray(integrations) || integrations.length === 0)
+    return undefined;
+
+  // We remove invalid integration items
+  const validIntegrations = integrations.filter(integration => {
+    return integration && validIntegrationTypes.includes(integration.type);
```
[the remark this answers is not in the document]
EmilianoSanchez: validIntegrationTypes is something we control, but, just in case, I will add that extra validation

src/utils/constants/index.js
EmilianoSanchez: Yes, like storage related constants.

src/trackers/impression.js
EmilianoSanchez: I would send the same shallow-copied object (with a reference to impression) to both integrations and impressionListener, since it is the current way used for impressionListener, and because for avoiding any unexpected change we should do a deep copy of the object, that is expensive.

PR #358, opened 2020-02-12
Split-to-Ga integration

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Split-to-GA integration module: it filters, maps and sends Split impressions and events as GA hits. This integration is only for Browser.

- E2E suite test for browser, using real GA tag but mocked axios for Split endpoints.
- [TODO] Unit tests of SplitToGa module.

## Extra Notes


src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
-    queue: function (data, type) {/* noop for now */ }
-  };
+import logFactory from '../../utils/logger';
+import { uniq } from '../../utils/lang';
+import { SPLIT_IMPRESSION, SPLIT_EVENT } from '../../utils/constants';
+const log = logFactory('splitio-split-to-ga');
+
+class SplitToGa {
+
+  static defaultFilter() { return true; }
```
NicoZelaya: Add some short and quick comments for this functions.

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+
+    this.filter = options && typeof (options.filter) === 'function' ?
+      options.filter :
+      SplitToGa.defaultFilter;
+
+    // @TODO Should we check something else about `configObject.impressionMapper`?
+    // It doesn't matter, because if the returned object is not a GA fieldsObject or string, ga send command will do nothing.
+    this.mapper = options && typeof (options.mapper) === 'function' ?
+      options.mapper :
+      SplitToGa.defaultMapper;
```
NicoZelaya: I think this assignment is not needed. It doesn't hurt either but you should get the static method anyways (that's why we make it static).

To do this, you'll need to rename the defaults to have the same names that the one we expect, "mapper" instead of "defaultMapper" and "filter" instead of "defaultFilter".

Another way to write these then would be (assuming you changed the options existance check already):
```
if (typeof options.mapper === 'function') this.mapper = options.mapper;
```

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+      log.warn(`SplitToGa queue method threw: ${err}. No hit was sent.`);
+    }
+  }
+
 }
 
-export default SplitToGaFactory;
\ No newline at end of file
+// A falsy object represents the default tracker
+SplitToGa.defaultTrackerNames = [''];
```
NicoZelaya: Why not to have this on the class definition?

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+
+    // Check if `ga` object is available
+    if (typeof SplitToGa.getGa() !== 'function') {
+      // @TODO review the following warning message
+      log.warn('`ga` command queue not found. No hits will be sent.');
+      // Return an empty object to avoid creating a SplitToGa instance
+      return {};
+    }
+
+    this.filter = options && typeof (options.filter) === 'function' ?
```
NicoZelaya: From this line going forward, all the statements on the constructor depend on the "options" variable being truthy. Use a top level if statement to check that, then in it you only check for the property of the options you care on each assignment.

src/integrations/ga/__tests__/SplitToGa.spec.js
```diff
@@ -0,0 +1,145 @@
+/**
+ * Unit tests:
+ *  DONE- SplitToGa.validateFieldsObject
+ *  DONE- SplitToGa.defaultFilter
+ *  DONE- SplitToGa.defaultMapper
+ *  DONE- SplitToGa.getGa
+ *  DONE- SplitToGa.prototype.constructor
+ *  DONE- SplitToGa.prototype.queue
+ */
```
NicoZelaya: If all done remove this.

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+      SplitToGa.defaultTrackerNames;
+  }
+
+  queue(data) {
+    try {
+      // filter and map Split events/impressions into a FieldsObject instance
+      const fieldsObject = this.filter(data) && this.mapper(data);
+
+      // don't send the hit if it is falsy or invalid when generated by a custom mapper
+      if (!fieldsObject || (this.mapper !== SplitToGa.defaultMapper && !SplitToGa.validateFieldsObject(fieldsObject)))
```
NicoZelaya: is the check for the mapper necessary?

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+      // We don't warn if a tracker does not exist, since the user might create it after the SDK is initialized.
+      // Note: GA allows to create and get trackers using a string or number as tracker name, and does nothing if other types are used.
+      uniq(options.trackerNames) :
+      SplitToGa.defaultTrackerNames;
+  }
+
+  queue(data) {
+    try {
+      // filter and map Split events/impressions into a FieldsObject instance
+      const fieldsObject = this.filter(data) && this.mapper(data);
```
NicoZelaya: If my understanding is correct, you're receiving here the data directly from the tracker. Either on the tracker or here it's better to have a copy of the object, as otherwise, if provided by reference, it could be modified.

Similarly here, you pass the reference to the filter and the same reference to the mapper. While we agree that it's cleaner and more convenient code-wise, provide a copy. Use spread or merge or whatever tool you need to use but let's try to protect our users as much as we can from themselves.

src/integrations/ga/SplitToGa.js
```diff
@@ -1,10 +1,108 @@
...
+    this.trackerNames = options && Array.isArray(options.trackerNames) ?
+      // We strip off duplicated values if we received a `trackerNames` param.
+      // We don't warn if a tracker does not exist, since the user might create it after the SDK is initialized.
+      // Note: GA allows to create and get trackers using a string or number as tracker name, and does nothing if other types are used.
+      uniq(options.trackerNames) :
+      SplitToGa.defaultTrackerNames;
+  }
+
+  queue(data) {
+    try {
```
NicoZelaya: Don't just wrap everything in this but only what's necessary. Unless this changed (it's been a couple of years since I looked into it) there are less optimizations possible on the machine code generated inside a try..catch block

types/splitio.d.ts
```diff
@@ -559,6 +571,39 @@ declare namespace SplitIO {
    */
   interface SplitToGaIntegration {
     type: 'SPLIT_TO_GA',
+    /**
+     * Optional filter to use instead of default, which always return true, 
+     * meaning that all impressions and events are tracked as GA hits. 
+     */
+    filter?: (data: SplitIO.Data) => boolean,
```
NicoZelaya: The name "SplitIO.Data" is too generic. This is not Split data as a whole. Should be IntegrationData or something like that.

types/splitio.d.ts
```diff
@@ -551,6 +560,9 @@ declare namespace SplitIO {
   interface GaToSplitIntegration {
     type: 'GA_TO_SPLIT',
   }
+  type SPLIT_IMPRESSION = 'IMPRESSION';
+  type SPLIT_EVENT = 'EVENT';
```
NicoZelaya: Do this ones need to be public/exposed? These will change when exposing the constants for clients use anyways, correct?

NicoZelaya: Unless we scope these we'll need to rename them or move them where they're not exposed too.

src/integrations/ga/SplitToGa.js
EmilianoSanchez: `static defaultTrackerNames = [''];` is not a valid JS syntax :/

src/integrations/ga/SplitToGa.js
EmilianoSanchez: Refactored, but I don't get why we should rename default methods.

src/integrations/ga/SplitToGa.js
EmilianoSanchez: yes, since we validate `fieldsObject` only if it was generated by a custom mapper (we assume default mapper always return a valid `fieldsObject`)

NicoZelaya: Why not always validating and having simpler code? instead of the assumption and the extra condition.

src/integrations/ga/SplitToGa.js
```diff
@@ -5,18 +5,15 @@ const log = logFactory('splitio-split-to-ga');
...
   static defaultMapper({ type, payload }) {
     switch (type) {
       case SPLIT_IMPRESSION:
         return {
           hitType: 'event',
           eventCategory: 'split-impression',
-          eventAction: payload.impression.feature,
-          eventLabel: payload.impression.treatment,
+          eventAction: 'Evaluate ' + payload.impression.feature,
+          eventLabel: 'Treatment ' + payload.impression.treatment + ' Label ' + payload.impression.label,
```
NicoZelaya: Let's make it: `'Treatment: ' + payload.impression.treatment + '. Targeting rule: ' + payload.impression.label + '.',`

PR #360, opened 2020-02-12
Ga-to-Split integration

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- GA-to-Split integration module: it filters, maps and sends GA hits as Split events. This integration is only for Browser, and is based on defining a custom GA plugin.

- E2E suite test for browser, using real GA tag but mocked axios for Split endpoints.
- [TODO] Unit tests of GaToSplit module.

## Extra Notes


src/utils/lang/index.js
```diff
@@ -157,6 +164,18 @@ export function uniq(arr) {
   });
 }
 
+/**
+ * Removes duplicate items on an array of objects using an optional `stringify` function as equality criteria.
+ * It uses JSON.stringify as default criteria.
+ */
+export function uniqO(arr, stringify = JSON.stringify) {
```
NicoZelaya: This name is not representative at all. either call it "uniqAsStrings" or just preprocess the items before passing them to the regular uniq.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+function providePlugin(pluginName, pluginConstructor) {
+  const gaAlias = window.GoogleAnalyticsObject || 'ga';
+  window[gaAlias] = window[gaAlias] || function (...args) {
+    (window[gaAlias].q = window[gaAlias].q || []).push(args);
+  };
+
+  // provides the plugin for use with analytics.js.
+  window[gaAlias]('provide', pluginName, pluginConstructor);
+
+  // @TODO should we registers the plugin on the global gaplugins object?
```
NicoZelaya: I would think so, what are the implications?

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+
+/**
+ * Provides a plugin to use with analytics.js, accounting for the possibility
+ * that the global command queue has been renamed or not yet defined.
+ * @param {string} pluginName The plugin name identifier.
+ * @param {Function} pluginConstructor The plugin constructor function.
+ */
+function providePlugin(pluginName, pluginConstructor) {
+  const gaAlias = window.GoogleAnalyticsObject || 'ga';
+  window[gaAlias] = window[gaAlias] || function (...args) {
```
NicoZelaya: Add a comment explaining why we have this.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+/**
+ * Build a mapper function based on a mapping object
+ *
+ * @param {object} mapping
+ */
+function mapperBuilder(mapping) {
+  return function (model) {
+    const hitType = model.get('hitType');
+
+    let eventTypeId = (mapping.eventTypeIdPrefix[hitType] || 'ga');
```
NicoZelaya: shouldn't prefix (non-default) come from the settings of the plugin?

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+  // window.gaplugins[capitalize(pluginName)] = pluginConstructor;
+}
+
+// Default filter: accepts all hits
+export function defaultFilter() { return true; }
+
+// Default mapping: object used for building the default mapper from hits to Split events
+// @TODO review default mapping.
+const defaultMapping = {
+  eventTypeIdPrefix: {
```
NicoZelaya: I don't know if prefix is the right name for these.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+
+  // Filter based on rum-agent identities validator
+  return uniqueIdentities.filter(identity => {
+    if (!identity)
+      return false;
+
+    const maybeKey = identity.key;
+    const maybeTT = identity.trafficType;
+
+    if (!isString(maybeKey) && !isNumber(maybeKey))
```
NicoZelaya: If not yet implemented, add a test to check what happens when you receive numbers.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+ * @returns {boolean} Whether the data instance is a valid EventData or not.
+ */
+export function validateEventData(data) {
+  if (!validateEvent(data.eventTypeId, 'splitio-ga-to-split:mapper'))
+    return false;
+
+  if (validateEventValue(data.value, 'splitio-ga-to-split:mapper') === false)
+    return false;
+
+  const { properties } = validateEventProperties(data.properties, 'splitio-ga-to-split:mapper');
```
NicoZelaya: Put "splitio-ga-to-split:mapper" in a const variable.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+
+      this.tracker = tracker;
+
+      // Validate identities
+      const validIdentities = validateIdentities(opts.identities);
+
+      if (!validIdentities || validIdentities.length === 0) {
+        log.warn('No valid identities were provided. Please check that you are passing a valid list of identities or providing a traffic type at the SDK configuration.');
+        return;
+      }
```
NicoZelaya: Where are you passing the "default identity" to this? (should only be there if identities is not defined by the user)

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+      undefined
+  };
+
+  class SplitTracker {
+
+    // Constructor for the SplitTracker plugin.
+    constructor(tracker, pluginOptions) {
+
+      // precedence of options: SDK options (config.integrations) overwrite pluginOptions (`ga('require', 'splitTracker', pluginOptions)`)
+      const opts = Object.assign({}, defaultOptions, sdkOptions, pluginOptions);
```
NicoZelaya: Assign the options to the instance.

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+        log.warn('The provided `prefix` was ignored since it is invalid. Please check that you are passing a string object as `prefix`.');
+        opts.prefix = undefined;
+      }
+
+      // Overwrite sendHitTask to perform plugin tasks:
+      // 1) filter hits
+      // 2) map hits to Split events
+      // 3) handle events, i.e., validate and send them to Split BE
+      const originalSendHitTask = tracker.get('sendHitTask');
+      tracker.set('sendHitTask', function (model) {
```
NicoZelaya: Maybe we could move this function body, but I'm not requesting this change.

Having options and tracker on the instance should make it possible. Might be another option to use another function to also overwrite the task so you keep the reference in that context, while accessing tracker and options from this. (will need to bind the context for the callback though).

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+
+        // Store the event
+        if (eventData.key && eventData.trafficTypeName) {
+          opts.eventHandler(event);
+        } else { // Store the event for each Key-TT pair (identities), if key and TT is not present in eventData
+          opts.identities.forEach(identity => {
+            const event = Object.assign({
+              key: identity.key,
+              trafficTypeName: identity.trafficType,
+            }, eventData);
```
NicoZelaya: If the event data were to contain only one of those values set, that would override the identity one. I guess that's expected right?

src/integrations/ga/GaToSplit.js
```diff
@@ -1,7 +1,257 @@
...
+ */
+function GaToSplit(sdkOptions, storage, coreSettings) {
+
+  const defaultOptions = {
+    filter: defaultFilter,
+    mapper: defaultMapper,
+    // Default eventHandler stores event
+    eventHandler: function (event) {
+      storage.events.track(event);
+    },
```
NicoZelaya: This seems odd, people will not define this in the options. But today if they define something with that name they might override this.

src/utils/lang/index.js
```diff
@@ -92,6 +92,13 @@ export function isString(val) {
   return typeof val === 'string' || val instanceof String;
 }
 
+/**
+ * Checks if a given value is a string.
+ */
```
NicoZelaya: Wrong comment

src/utils/lang/index.js
```diff
@@ -92,6 +92,13 @@ export function isString(val) {
...
+export function isNumber(val) {
+  return typeof val === 'number' || val instanceof Number;
+}
+
```
NicoZelaya: Why not using isFinite? It's the same validation we use somewhere else as probably having a number that results in NaN or Infinity is not expected by our customers but rather an issue on a calculation.

src/utils/lang/index.js
EmilianoSanchez: Renamed. By passing the preprocessed items to regular uniq, we lose the reference of what original items were removed.

src/integrations/ga/GaToSplit.js
EmilianoSanchez: Removed. autotrack uses it for testing purposes.

src/integrations/ga/GaToSplit.js
EmilianoSanchez: Yes

src/integrations/ga/GaToSplit.js
EmilianoSanchez: Removed to simplify default mapper and since a prefix can be passed as options ('ga' by default)

src/integrations/ga/GaToSplit.js
```diff
@@ -130,34 +111,56 @@ export function validateIdentities(identities) {
...
+  // set a default eventTypeId if it is not a string or is an empty one.
+  if (!isString(eventTypeId) || eventTypeId.length === 0) {
+    return DEFAULT_EVENT_TYPE;
+  }
+
+  // replace invalid substrings and truncate
+  const fixed = eventTypeId
+    .replace(INVALID_PREFIX_REGEX, '')
+    .replace(INVALID_SUBSTRING_REGEX, '_')
+    .slice(0, 80);
```
NicoZelaya: why 80?

EmilianoSanchez: part of EVENT_TYPE_REGEX: no more than 80 chars.

PR #361, opened 2020-02-18
Telemetry noise

NicoZelaya:
# JS SDK

## What did you accomplish?
Reduced the noise caused by telemetry errors.

Download the code, bundle & transpile and test with regular setup. Then break the URL (via the config) for events and compare error logs coming from events & impressions failing to post vs telemetry error logs.

## Extra Notes
Ticket #sdks-1342

### DIFF
--- a/src/utils/settings/index.js
+++ b/src/utils/settings/index.js
@@ -49,7 +49,7 @@ const base = {
     // fetch segments updates each 60 sec
     segmentsRefreshRate: 60,
     // publish metrics each 60 sec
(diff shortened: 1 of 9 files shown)


src/utils/settings/index.js
```diff
@@ -49,7 +49,7 @@ const base = {
     // fetch segments updates each 60 sec
     segmentsRefreshRate: 60,
     // publish metrics each 60 sec
```
EmilianoSanchez: the comment says 60

PR #365, opened 2020-02-27
POC: only exposing getTreatmentsWithConfig and its dependencies

rohanmurthy:
GoDaddy has a case where we want to keep the bundle size of this package as small as possible. The relevant code to GoDaddy is for decisioning shoppers/visitors into treatments; essentially just `getTreatmentsWithConfig` (and `./src/engine/evaluator`). We were thinking of just exposing this logic as a separate package. This would require us to pass in a `storage` object to `getTreatmentsWithConfig`; we would also want to expose `SplitCacheLocalStorage, KeyBuilderLocalStorage, and SettingsFactory` as a part of this package to create this storage object properly.

Also, for `getTreatments`, we don't need any of the tracking logic. The main requested logic is reflected in the `src-new-package/client/client.js` file of this PR.


chillaq:
Hi Rohan, checking internally on this request.

Thanks
Bilal

rohanmurthy:
Hey @chillaq ! Thanks for the response! We have been in contact with @NicoZelaya about this. This is just a PR to illustrate what we wanted out of a potential feature we were thinking about.

rohanmurthy:
Circled back with the folks on my team, we are going to close this PR for now. Will potentially follow up later on this!

Arditc:
Hey, @rohanmurthy I really like what you've done there if honest and I was wondering with the PoC which you've provided did you see any issues with it?

PR #366, opened 2020-02-28
[DO NOT MERGE] Spike Adding Cloudflare Workers Support

RaniSputnik:
# JS SDK

## What did you accomplish?

In an ideal world, we would like to run Split in [consumer mode](https://help.split.io/hc/en-us/articles/360020564931-Node-js-SDK#consumer-mode) on [Cloudflare Workers](https://developers.cloudflare.com/workers/). This is a dirty spike to understand what changes would be required.

Don't! This is throwaway code.

## Extra Notes

The difficulty of running Split on a CloudFlare worker is born out of the fact that CloudFlare workers take what is traditionally a client side API (service workers) and run this server side. The SDK bakes in the assumption that if you are in a non-node environment, then you are client side which is not true for our use case.

* [This commit](https://github.com/splitio/javascript-client/commit/690ea721893b4ce203aa2bdf852f7f88d5871495) gets the Split SDK running on a CloudFlare worker in "browser" mode - we replace [axios](https://github.com/axios/axios) to remove the dependency on `XMLHTTPRequest` (CloudFlare workers use the fetch API).
* [This commit](https://github.com/splitio/javascript-client/commit/ba81eeeed6967305627b9d79df2312001431c6b0) messes with a bunch of `package.json` files to force webpack to compile the Node.js SDK for the browser. This allows us to instantiate a client without being forced to provide a "key" on start-up.
* The rest of this MR is then spent experimenting with which storage methods we need to implement to run the CloudFlare Worker in "consumer" mode, pointing at [Workers KV](https://developers.cloudflare.com/workers/reference/storage/).

### DIFF
--- a/src/client/inputValidation.js
+++ b/src/client/inputValidation.js
@@ -12,7 +12,7 @@ import {
   validateIfReady
 } from '../utils/inputValidation';
 import { startsWith } from '../utils/lang';
-import { STORAGE_REDIS, CONTROL, CONTROL_WITH_CONFIG } from '../utils/constants';
+import { STORAGE_REDIS, CONTROL, CONTROL_WITH_CONFIG, STORAGE_CLOUDFLARE_KV } from '../utils/constants';
 
 /**
  * We will validate the input before actually executing the client methods. We should "guard" the client here,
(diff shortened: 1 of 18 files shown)


NicoZelaya:
Closing as this was mostly just for reference

PR #382, opened 2020-03-24
[Push] Splits, Segments and MySegments Synchronizers

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- `SplitUpdateWorker`: module for queueing and handling SPLIT_UPDATE and SPLIT_KILL notifications.
- `SegmentUpdateWorker`: module for queueing and handling SEGMENT_UPDATE notifications.
- `MySegmentUpdateWorker`: module for queueing and handling MY_SEGMENTS_UPDATE notifications.
- New functions for storage, producers and updaters to support the new PUSH mode:
  - `killLocally` function used to kill Split at storage
  - `synchronizeSplits`, `synchronizeSegment`, and `synchronizeMySegments` functions in producers
  - Updated signature of `SegmentChangesUpdater` and `MySegmentChangesUpdater` to accepts a given segment name or list of segments respectively.

- unit tests for new modules

## Extra Notes


src/producer/browser/Full.js
```diff
@@ -29,30 +29,66 @@ const FullBrowserProducer = (context) => {
...
+  let isSynchronizeSplitsRunning = false;
+
+  function synchronizeSplits() {
+    isSynchronizeSplitsRunning = true;
+    return splitsUpdater().finally(function() {
+      isSynchronizeSplitsRunning = false;
+    });
+  }
+
+  let isSynchronizeMySegmentsRunning = false;
```
NicoZelaya, line 41: Define all vars together.

src/producer/browser/Full.js
```diff
@@ -29,30 +29,66 @@ const FullBrowserProducer = (context) => {
   const splitsUpdater = SplitChangesUpdater(context);
   const segmentsUpdater = MySegmentsUpdater(context);
 
+  let isSynchronizeSplitsRunning = false;
```
NicoZelaya, line 32: these vars are weird.. don't base them on the name of the method you use today but rather the action of synchronizing Splits. This is a boolean flag that indicates if the synchronization of splits is running, is doesn't actually mean the method below (synchronizeSplits) is currently being executed.

isSynchronyzingSplits and isSynchronizingSegments ?

src/producer/browser/Full.js
```diff
@@ -29,30 +29,66 @@ const FullBrowserProducer = (context) => {
...
+    return splitsUpdater().finally(function() {
+      isSynchronizeSplitsRunning = false;
+    });
+  }
+
+  let isSynchronizeMySegmentsRunning = false;
+
+  function synchronizeMySegments(segmentList) {
+    isSynchronizeMySegmentsRunning = true;
+    return segmentsUpdater(undefined, segmentList).finally(function () {
```
NicoZelaya, line 45: If you want no retries, use 0 instead of undefined. The default assignment of the variable might change, specially in a world where we modularize.

src/producer/browser/Full.js
```diff
@@ -29,30 +29,66 @@ const FullBrowserProducer = (context) => {
...
+    });
+  }
+
   const settings = context.get(context.constants.SETTINGS);
   const { splits: splitsEventEmitter } = context.get(context.constants.READINESS);
-  
-  const splitsUpdaterTask = TaskFactory(splitsUpdater, settings.scheduler.featuresRefreshRate);
+
+  const splitsUpdaterTask = TaskFactory(synchronizeSplits, settings.scheduler.featuresRefreshRate);
   const segmentsUpdaterTask = TaskFactory(segmentsUpdater, settings.scheduler.segmentsRefreshRate);
```
NicoZelaya, line 54: Shouldn't this be using synchronizeMySegments ? Why only splits use the wrapping function but not the segments?

src/producer/browser/Partial.js
```diff
@@ -24,15 +24,36 @@ import onSplitsArrivedFactory from './onSplitsArrivedFactory';
...
   const segmentsUpdater = MySegmentsUpdater(context);
   const segmentsUpdaterTask = TaskFactory(segmentsUpdater, settings.scheduler.segmentsRefreshRate);
 
   const onSplitsArrived = onSplitsArrivedFactory(segmentsUpdaterTask, context);
-  
+
   splitsEventEmitter.on(splitsEventEmitter.SDK_SPLITS_ARRIVED, onSplitsArrived);
 
-  return segmentsUpdaterTask;
+  let isSynchronizeMySegmentsRunning = false;
```
NicoZelaya, line 35: also suggest you change this name.

src/producer/browser/Partial.js
```diff
@@ -24,15 +24,36 @@ import onSplitsArrivedFactory from './onSplitsArrivedFactory';
...
+
+  function synchronizeMySegments(segmentList) {
```
NicoZelaya, line 37: Shouldn't you assign this to the task so you can keep track of those updates too? Or should you remove the splits one from the full producer splits task?

src/producer/updater/MySegments.js
```diff
@@ -32,24 +32,34 @@ function MySegmentsUpdaterFactory(context) {
...
+      readyOnAlreadyExistentState = false;
+      segmentsEventEmitter.emit(segmentsEventEmitter.SDK_SEGMENTS_ARRIVED);
+    }
+  }
+
+  return function MySegmentsUpdater(retry = 0, segmentList) {
+    // If segmentList is provided, there is no need to fetch mySegments
+    if (segmentList) {
+      updateSegments(segmentList);
+      return;
```
NicoZelaya, line 50: You should also return a promise on this path, otherwise you have two different types to be expected.

src/producer/node.js
```diff
@@ -31,21 +31,49 @@ const NodeUpdater = (context) => {
   let stopSplitsUpdate = false;
   let stopSegmentsUpdate = false;
   let splitFetchCompleted = false;
+  let isRunning = false;
+  let isSynchronizeSplitsRunning = false;
+  let isSynchronizeSegmentRunning = false;
```
NicoZelaya, line 36: Same as before.

src/producer/node.js
```diff
@@ -59,26 +87,48 @@ const NodeUpdater = (context) => {
...
+    isSynchronizeSplitsRunning() {
+      return isSynchronizeSplitsRunning;
+    },
+    synchronizeSplits,
+
+    // Used by SegmentUpdateWorker
+    isSynchronizeSegmentRunning() {
+      return isSynchronizeSegmentRunning;
+    },
+    synchronizeSegment,
```
NicoZelaya, line 131: trailing comma

src/storage/SplitCache/killLocally.js
```diff
@@ -0,0 +1,24 @@
+/**
```
NicoZelaya, line 1: Why is this not part of the storage? Either we make it part of the Split caches (where each one requires it's own handling) or we take this functionality, ideally, at the same place where we're executing this and passing the storage down.

The way it is today it's tricky because you would need to handle the asynchronicity of the getSplit and the addSplit methods here, would be better if we move it to each storage (you're taking advantage of the fact that the only async storage we have today is Redis where we don't sync the data, but that's changing in the next 6 months)

src/sync/SegmentUpdateWorker/SegmentUpdateWorker.js
```diff
@@ -0,0 +1,50 @@
...
+   * @param {string} segmentName segment name of the SEGMENT_UPDATE notification
+   */
+  put(changeNumber, segmentName) {
+    const currentChangeNumber = this.segmentsStorage.getChangeNumber(segmentName);
+
+    if (changeNumber <= currentChangeNumber) return;
+
+    this.segmentsChangesQueue.push({ segmentName, changeNumber });
+
+    if (this.segmentsProducer.isSynchronizeSegmentRunning()) return;
```
NicoZelaya, line 45: This is odd. You can sync few segments in parallel and there's no reason not to update segment A at the same time than segment B, it just doesn't make sense to trigger two requests at the same time for segment B.

src/producer/browser/Full.js
EmilianoSanchez: changed to `isSynchronyzingSplits`, `isSynchronizing[My]Segments`

src/producer/browser/Full.js
EmilianoSanchez: yes!, actually I found a bug, thanks to that comment

src/producer/browser/Partial.js
EmilianoSanchez: done

src/storage/SplitCache/killLocally.js
EmilianoSanchez: refactored: killLocally was added to the prototypes of Split caches.

src/sync/SegmentUpdateWorker/SegmentUpdateWorker.js
EmilianoSanchez: Done. Now `SegmentUpdateWorker` keeps a map of segment names to `maxChangeNumbers`, instead of a `segmentsChangesQueue` (similar approach followed by `SplitUpdateWorker` and `MySegmentUpdateWorker`)

src/utils/fn/repeat.js
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 **/
 
-function repeat(fn, delay, ...rest) {
+function repeat(fn, delay, notCallImmediately, ...rest) {
```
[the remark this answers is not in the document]
EmilianoSanchez: I would prefer to delay this refactor for the final polishing, since it is not as trivial as it seems at first glance, and I am not seeing a real improvement in the code.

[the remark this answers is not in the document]
NicoZelaya: Out of curiosity, what's the criteria you're using to meassure the code improvement? maintainability? amount of lines?

PR #383, opened 2020-03-24
[Push] SSEHandler, including NotificationProcessor, Parser and Keeper.

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- `SSEHandler` (SSEHandler/index.js)
- `NotificationProcessor` (SSEHandler/index.js::handleMessage)
- `NotificationParser` (SSEHandler/NotificationParser.js)
- `NotificationManagerKeeper` (SSEHandler/NotificationKeeper.js)

- Unit tests

## Extra Notes


src/sync/NotificationProcessor/index.js
```diff
@@ -0,0 +1,91 @@
+import { errorParser, messageParser } from './notificationparser';
+import Backoff from '../../utils/backoff';
+import { PushEventTypes } from '../constants';
+
+const controlPriMatcher = /control_pri$/;
+
+// @TODO logging
+export default function NotificationProcessorFactory(
+  sseClient,
+  pushEmitter, // feedbackLoop
```
EmilianoSanchez, line 10: queues

src/sync/SSEHandler/index.js
```diff
@@ -0,0 +1,86 @@
...
+            parsedData.defaultTreatment);
+          break;
+
+        /* occupancy & control events, handled by NotificationManagerKeeper */
+        case OCCUPANCY:
+          notificationKeeper.handleOccupancyEvent(parsedData.metrics.publishers, channel, timestamp);
+          break;
+
+        case CONTROL:
+          notificationKeeper.handleControlEvent(parsedData.controlType, channel, timestamp);
```
NicoZelaya, line 80: never miss a break on a switch.

src/sync/SSEHandler/index.js
```diff
@@ -0,0 +1,86 @@
...
+        return;
+      }
+
+      const { parsedData, channel, timestamp } = messageWithParsedData;
+
+      // we only handle update events if streaming is up.
+      if (!notificationKeeper.isStreamingUp() && parsedData.type !== OCCUPANCY && parsedData.type !== CONTROL)
+        return;
+
+      switch (parsedData.type) {
```
NicoZelaya, line 48: Add the default clause at the end.

src/sync/SSEHandler/index.js
```diff
@@ -0,0 +1,86 @@
...
+          break;
+
+        case MY_SEGMENTS_UPDATE:
+          pushEmitter.emit(MY_SEGMENTS_UPDATE,
+            parsedData,
+            channel);
+          break;
+
+        case SPLIT_KILL:
+          pushEmitter.emit(SPLIT_KILL,
```
NicoZelaya, line 68: Consider avoiding the need for as many new lines on the update handling.

src/sync/SSEHandler/NotificationKeeper.js
```diff
@@ -0,0 +1,56 @@
+import { PUSH_CONNECT, PUSH_DISCONNECT, PUSH_DISABLED, ControlTypes } from '../constants';
+
+const controlPriMatcher = /control_pri$/;
```
NicoZelaya, line 3: Similar to the one in SSE client. Also we should have other regexes so maybe follow that name convention? 

Given it's constant, should be all capitalized letters.

src/sync/SSEHandler/NotificationKeeper.js
```diff
@@ -0,0 +1,56 @@
...
+
+export default function notificationKeeperFactory(feedbackLoopEmitter) {
+
+  let isStreamingUp;
```
NicoZelaya, line 7: Start it with false. It's usually not great to change the type of the variable (reduces applicable machine code optimizations besides making it more confusing), even if JS allows for it. 
Use undefined for when you need to check if something is not set/defined, not only as a falsey value. For falsiness go with the boolean `false`.

src/sync/SSEHandler/NotificationKeeper.js
```diff
@@ -0,0 +1,56 @@
...
+      feedbackLoopEmitter.emit(PUSH_CONNECT);
+    },
+
+    handleOccupancyEvent(publishers, channel, timestamp) {
+      if (controlPriMatcher.test(channel) && timestamp > occupancyTimestamp) {
+        occupancyTimestamp = timestamp;
+        if (publishers === 0 && isStreamingUp) {
+          isStreamingUp = false;
+          feedbackLoopEmitter.emit(PUSH_DISCONNECT); // notify(STREAMING_DOWN) in spec
+          return;
```
NicoZelaya, line 23: If you use an else before the next if (keep the if too as I think it's needed) you don't need to have this return. I don't know if the sonarqube configured today tracks this as an increase in complexity but still, easy to avoid. 

FYI only asking this as there's nothing else following the if on line 25. If the function execution would continue further then the return is totally fine.

src/sync/SSEHandler/NotificationKeeper.js
```diff
@@ -0,0 +1,56 @@
...
+        }
+        if (controlType === ControlTypes.STREAMING_RESUMED && !isStreamingUp) {
+          isStreamingUp = true;
+          feedbackLoopEmitter.emit(PUSH_CONNECT); // notify(STREAMING_UP) in spec
+          return;
+        }
+        if (controlType === ControlTypes.STREAMING_DISABLED) {
+          isStreamingUp = false;
+          feedbackLoopEmitter.emit(PUSH_DISABLED);
```
NicoZelaya, line 47: Consider using else ifs to control the flow. Try not to think of returns as a GOTO, but a way to break execution. If all you need is for stopping after one condition was met, use a switch or if elseif

_If you love your returns, don't just leave it out because the current code works with that. If later someone else comes, adds a new type and misses the lack of return (and don't test properly of course), a bug might be introduced._

src/sync/constants.js
```diff
@@ -0,0 +1,24 @@
+// time for refresh token
+export const SECONDS_BEFORE_EXPIRATION = 600;
+
+// Internal SDK events, subscribed by SyncManager and PushManager
+export const PUSH_CONNECT = 'PUSH_CONNECT';
+export const PUSH_DISCONNECT = 'PUSH_DISCONNECT';
+export const SSE_ERROR = 'SSE_ERROR';
+export const PUSH_DISABLED = 'PUSH_DISABLED';
```
NicoZelaya, line 8: (applies to the file itself)

Keeping related constants close together makes total sense, but apply a criteria there too. Alphabetic works, priority if clear works. Anything works as long as there's a criteria that other people can understand.

src/sync/AuthClient/index.js
```diff
@@ -7,17 +7,19 @@ import { decodeJWTtoken } from '../../utils/jwt';
...
     .then(resp => resp.data)
     .then(json => {
-      return {
-        ...json,
-        decodedToken: decodeJWTtoken(json.token),
-      };
+      if (json.token) // empty token when `"pushEnabled": false`
+        return {
+          ...json,
+          decodedToken: decodeJWTtoken(json.token),
```
NicoZelaya, line 21: traling comma.

src/sync/AuthClient/index.js
```diff
@@ -7,17 +7,19 @@ import { decodeJWTtoken } from '../../utils/jwt';
...
+        };
+      return json;
```
NicoZelaya, line 23: This one deserves some { }

src/sync/SSEHandler/index.js
```diff
@@ -0,0 +1,86 @@
...
+      notificationKeeper.handleOpen();
+    },
+
+    /* HTTP & Network errors */
+    handleError(error) {
+      let errorWithParsedData = error;
+      try {
+        errorWithParsedData = errorParser(error);
+      } catch (err) {
+        log.error(`Error parsing SSE error notification: ${err}`);
```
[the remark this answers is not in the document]
EmilianoSanchez: yes. it shouldn't be logged as an error. I changed it to warning.

src/sync/SSEHandler/NotificationKeeper.js
EmilianoSanchez: changed these new regex to follow the naming convention: X_CHANNEL_REGEX

PR #384, opened 2020-03-24
[Push] PushManager, including Feedbackloop and Backoff algorithm

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- `PushManager`: handles initialization of push mode, [re-]authentications, and provides an eventEmitter to manage feedback loop and Push update notifications.
- `Backoff` algorithm

- unit tests [DOING PushManager tests]

## Extra Notes


src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+    const userKeys = clientContexts ? Object.keys(clientContexts) : undefined;
+    authenticate(settings, userKeys).then(
+      function (authData) {
+        // restart backoff retry counter for auth and SSE connections, due to HTTP/network errors
+        reauthBackoff.reset();
+        sseReconnectBackoff.reset(); // reset backoff in case SSE conexion has opened after a HTTP or network error.
+
+        // emit PUSH_DISCONNECT if org is not whitelisted
+        if (!authData.pushEnabled) {
+          log.error('Streaming is not enabled for the organization. Switching to polling mode.');
```
NicoZelaya, line 68: Use a comment in the code to mark it for review before releasing.

For now, use something like: "Streaming is not available. Switching to polling mode."

don't use error level, because it'll create lots of noise during the rollout phases. Go with INFO.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+        // Branch for other HTTP and network errors
+        const delayInMillis = reauthBackoff.scheduleCall();
+        log.error(`Fail to authenticate for push notifications, with error message: "${error.message}". Attempting to reauthenticate in ${delayInMillis / 1000} seconds.`);
+      }
+    );
+  }
+
+  function disconnectPush() {
+    sseClient.close();
+
```
NicoZelaya, line 106: Log an info level statement, to mark the disconnection. Would use another one for marking the connection.

Disregard if you're already doing it somewhere else.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+  });
+
+  /** Fallbacking due to SSE errors */
+
+  pushEmitter.on(SSE_ERROR, function (error) { // HTTP or network error in SSE connection
+    // SSE connection is closed to avoid repeated errors due to retries
+    sseClient.close();
+
+    // retries are handled via backoff algorithm
+    let delayInMillis = (error.parsedData && (error.parsedData.statusCode === 400 || error.parsedData.statusCode === 401)) ?
```
NicoZelaya, line 127: Assign the code to a variable to make the ternary more readable. If you want to keep it as is, with long conditions might be best to use an if statement directly so the two flows are clear.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+
+  // map of hashes to user keys, to dispatch MY_SEGMENTS_UPDATE events to the corresponding MySegmentsUpdateWorker
+  const userKeyHashes = {};
+  // variable used on browser to reconnect only when a new client was added, saving some authentication and sse connections.
+  let connectForNewClient = false;
+
+  /** PushManager functions related to initialization */
+
+  const reauthBackoff = new Backoff(connectPush, settings.authRetryBackoffBase);
+  const sseReconnectBackoff = new Backoff(sseClient.reopen.bind(sseClient), settings.streamingReconnectBackoffBase);
```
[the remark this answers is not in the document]
NicoZelaya: Already discussed on Slack. Will check some things and try to put the binding in the class constructor itself.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+    pushEmitter.on(SEGMENT_UPDATE, segmentUpdateWorker.put.bind(segmentUpdateWorker));
+  }
+
+  return Object.assign(
+    // Expose Event Emitter functionality and Event constants
+    Object.create(pushEmitter),
+    {
+
+      // Expose functionality for starting and stoping push mode:
+
```
NicoZelaya, line 167: Don't abuse empty lines unnecessarily.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+      // used in browser
+      startNewClient(userKey, context) {
+        const hash = hashUserKey(userKey);
+        if (!userKeyHashes[hash]) {
+          userKeyHashes[hash] = userKey;
+          connectForNewClient = true; // we must reconnect on start, to listen the channel for the new user key
+        }
+        const storage = context.get(context.constants.STORAGE);
+        const producer = context.get(context.constants.PRODUCER);
```
NicoZelaya, line 181: I think you could safely declare these after the has constant. Apply some code styling as we discussed previously.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+
+        // Reconnects in case of a new client.
+        // Run in next event-loop cycle to save authentication calls
+        // in case the user is creating several clients in the current cycle.
+        setTimeout(function start() {
+          if (connectForNewClient) {
+            connectForNewClient = false;
+            connectPush();
+          }
+        });
```
NicoZelaya, line 192: Always add the 0, to make it clearer to the unexperienced eye.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,201 @@
...
+          connectForNewClient = true; // we must reconnect on start, to listen the channel for the new user key
+        }
+        const storage = context.get(context.constants.STORAGE);
+        const producer = context.get(context.constants.PRODUCER);
+        context.put(context.constants.MY_SEGMENTS_CHANGE_WORKER, new SegmentUpdateWorker(storage.segments, producer));
+
+        // Reconnects in case of a new client.
+        // Run in next event-loop cycle to save authentication calls
+        // in case the user is creating several clients in the current cycle.
+        setTimeout(function start() {
```
NicoZelaya, line 187: If you give it name for debugging (which is good, as it'll show up on stack traces) make it a bit clearer. maybe checkForReconnect ?

src/sync/PushManager/index.js
NicoZelaya: Emi, let's please make sure we change this before merging.

src/sync/PushManager/index.js
NicoZelaya: Same, let's make sure (if you already did disregard but haven't seen a reaction to this comment indicating it)

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,202 @@
...
+        scheduleTokenRefresh(decodedToken.iat, decodedToken.exp);
+      }
+    ).catch(
+      function (error) {
+
+        sseClient.close(); // no harm if already disconnected
+        pushEmitter.emit(PUSH_DISCONNECT); // no harm if `PUSH_DISCONNECT` was already notified
+
+        if (error.statusCode) {
+          switch (error.statusCode) {
```
NicoZelaya, line 89: I think we'll only have this error code with a special handling here correct? let's just use an if in that case.

src/sync/PushManager/index.js
```diff
@@ -0,0 +1,202 @@
...
+        scheduleTokenRefresh(decodedToken.iat, decodedToken.exp);
+      }
+    ).catch(
+      function (error) {
+
+        sseClient.close(); // no harm if already disconnected
+        pushEmitter.emit(PUSH_DISCONNECT); // no harm if `PUSH_DISCONNECT` was already notified
+
+        // Handle HTTP error for invalid API Key
+        if (error.statusCode && error.statusCode === 401) {
```
NicoZelaya, line 90: ping

PR #386, opened 2020-03-30
[Push] SyncManager and new config options

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- `SyncManager`
- Config options: `streamingEnabled`, `authRetryBackoffBase`, `streamingReconnectBackoffBase`, new urls.
- Update `SplitFactory`
- Other fixes:
  - Update GA type declaration comments.
  - Bug fix: destroying the main client must clean the periodic fetch (MySegments) of all clients. destroying shared clients must only clean their periodic fetch.

- [TODO] unit tests

## Extra Notes


src/utils/settings/index.js
```diff
@@ -139,6 +148,15 @@ function defaults(custom) {
   // `integrations` returns an array of valid integration items.
   withDefaults.integrations = integrations(withDefaults);
 
+  // validate push options
+  if (typeof withDefaults.streamingEnabled !== 'boolean') withDefaults.streamingEnabled = false;
```
NicoZelaya, line 152: Remember you have utilities to use, you should aim to use them as much as possible. Having a single definition of what something is (in this case) allows for easier and faster changes whenever we adapt our rules or want to apply a quick performance optimization.

src/utils/settings/index.js
```diff
@@ -139,6 +148,15 @@ function defaults(custom) {
...
+  if (withDefaults.streamingEnabled === true) {
```
NicoZelaya, line 153: By this line it's a boolean already (if it wasn't, you assigned false on the previous line). Just go with `if (withDefaults.streamingEnabled)`

src/utils/settings/index.js
```diff
@@ -90,6 +90,15 @@ const base = {
...
   integrations: undefined,
+
+  // toggle using (true) or not using (false) Server-Side Events for synchronizing storage
+  streamingEnabled: false,
+
+  // backoff base seconds to wait before re attempting to authenticate for push notifications
+  authRetryBackoffBase: 1,
+
+  // backoff base seconds to wait before re attempting to connect to streaming
+  streamingReconnectBackoffBase: 1,
```
NicoZelaya, line 101: trailing comma.

src/sync/node.js
```diff
@@ -0,0 +1,65 @@
+import PushManagerFactory from './PushManager';
+import { PUSH_DISCONNECT, PUSH_CONNECT } from './constants';
+import FullProducerFactory from '../producer';
+import logFactory from '../utils/logger';
+const log = logFactory('splitio-sync:sync-manager');
```
NicoZelaya, line 5: Might be too manija but try ordering the top section by following a criteria and repeating across all files (not just the ones that mirror each other). For example you could do:
Factories
utilities
CONSTANTS
variables/instances

If you are trying to group by semantics, there is either something wrong with the placement (if the constants are shared, don't put them between factories) 
or with the placement of that constants file (as in it's not for the whole sync but only PUSH hence you have them close, move to the push folder maybe so it's easy to extract as a unit, consider re exporting through PUSH manager if that's the case). 
Consider using spaces or comments if grouping by semantic/functionality, but would be better to stay away of the categories might not always be clear.

Apply this comment to all files here and in other PRs, not necessarily right now (but it's harmless and better to get it off the way) but ideally before closing development.

src/sync/node.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  function syncAll() {
+    // fetch splits and segments. There is no need to catch this promise (it is handled by `SplitChangesUpdater`)
+    producer.synchronizeSplits().then(() => {
+      producer.synchronizeSegment();
+    });
+  }
+
+  let pushManager;
```
NicoZelaya, line 38: If you could reorder variables and function declarations too, that's better.

src/sync/node.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const settings = context.get(context.constants.SETTINGS);
+  if (settings.streamingEnabled)
```
NicoZelaya, line 41: Consider marking the block ({}) if you're adding the line jumps next to the ifs everywhere. Or if you want use them as one liners.

src/sync/browser.js
```diff
@@ -0,0 +1,105 @@
...
+ */
+export default function BrowserSyncManagerFactory(mainContext) {
+
+  const contexts = {};
+  let pushManager;
+
+  function startPolling() {
+    log.info('PUSH down or disconnected. Starting periodic fetch of data.');
+    forOwn(contexts, function (context) {
+      const producer = context.get(context.constants.PRODUCER, true);
```
NicoZelaya, line 23: Do you expect to call this without an actual producer on the context or just being cautious? When do we expect it not to have a producer? 
Can't we use the context returned promise so we kick off polling as soon as the producer is assigned? Or what would happen if you called for start polling but you had no producer and don't schedule using the promise?



Question applies to all gets here, so if one is different, I'd be interesting in knowing it as well.

src/sync/browser.js
```diff
@@ -0,0 +1,105 @@
...
+    // fetch splits and segments
+    const mainProducer = mainContext.get(mainContext.constants.PRODUCER, true);
+    mainProducer && mainProducer.synchronizeSplits();
+    forOwn(contexts, function (context) {
+      const producer = context.get(context.constants.PRODUCER, true);
+      producer && producer.synchronizeMySegments();
+    });
+  }
+
+  function shared(context, isSharedClient = true) {
```
NicoZelaya, line 50: This should not be called shared, also the boolean flag is not necessary, specially given the form of the assignment in line #103. 
Invert the parameters, to create the instance you return as main one just call it with (false, context). Then in #103 do (using an example name) `syncManager.shared = createInstance.bind(null, true)` and just call the .shared() method from outside. 

Using a "shared" that creates the full instance and then reexporting as shared property on the instance itself, forcing a boolean telling IT IS a shared instance even though the property accessor is "shared"... seems like entangled.

src/sync/browser.js
```diff
@@ -0,0 +1,105 @@
...
+          if (producer && producer.isRunning())
+            producer.stop();
+
+        }
+      }
+    };
+  }
+
+  // called before creating PushManager, to create the producer and put in context.
+  const result = shared(mainContext, false);
```
NicoZelaya, line 96: "result" is not representative at all. why not just syncManager or mainSyncManager ?

src/sync/browser.js
```diff
@@ -0,0 +1,105 @@
...
+      producer && producer.synchronizeMySegments();
+    });
+  }
+
+  function shared(context, isSharedClient = true) {
+    const producer = isSharedClient ? PartialProducerFactory(context) : FullProducerFactory(context);
+    context.put(context.constants.PRODUCER, producer);
+    const settings = context.get(context.constants.SETTINGS);
+    const userKey = matching(settings.core.key);
+    if(contexts[userKey]) log.warn('A client with the same user key has already been created. Only the new instance will be properly synchronized.');
```
NicoZelaya, line 55: keep your spacing consistent. Both in line jumps and space before the parenthesis. Let's schedule a task later to review the linting and adjust some light autofix rules at some point of Q1.

src/sync/browser.js
```diff
@@ -0,0 +1,105 @@
...
+    mainProducer && mainProducer.synchronizeSplits();
+    forOwn(contexts, function (context) {
+      const producer = context.get(context.constants.PRODUCER, true);
+      producer && producer.synchronizeMySegments();
+    });
+  }
+
+  function shared(context, isSharedClient = true) {
+    const producer = isSharedClient ? PartialProducerFactory(context) : FullProducerFactory(context);
+    context.put(context.constants.PRODUCER, producer);
```
NicoZelaya, line 52: Put the variables towards the top and assignments after unless it makes sense execution~wise or you're defining blocks using spaces/comments. Any of those work.

src/utils/context/constants.js
```diff
@@ -8,7 +8,9 @@ export const SETTINGS = 'settings';
 export const STATUS_MANAGER = 'status_manager';
 export const STORAGE = 'storage';
 export const INTEGRATIONS_MANAGER = 'integrations_manager';
+export const PRODUCER = 'producer';
+export const MY_SEGMENTS_CHANGE_WORKER = 'my_segments_change_worker';
```
NicoZelaya, line 12: These were all in alphabetic order (I missed it when reviewing the integrations one). Please keep them ordered.

src/sync/browser.js
NicoZelaya: I think we have this pattern in other pieces of the code so let's have a chat about it tomorrow to make sure we're properly in sync about what and specially why I'm asking.

types/splitio.d.ts
```diff
@@ -809,7 +835,27 @@ declare namespace SplitIO {
...
+     * @default 1
+     */
+    authRetryBackoffBase?: number,
+    /**
+     * Seconds to wait before re attempting to connect to streaming.
+     * Next attempts follow intervals in power of two: base seconds, base x 2 seconds, base x 4 seconds, ...
+     * @property {number} streamingReconnectBackoffBase
+     * @default 1
+     */
+    streamingReconnectBackoffBase?: number,
```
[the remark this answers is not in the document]
EmilianoSanchez: the problem is that we don't share these properties with `INodeAsyncSettings`

src/sync/node.js
EmilianoSanchez: thanks. it is good to follow that import order convention

src/sync/browser.js
EmilianoSanchez: There is always a producer in all contexts when calling `startPolling`, `stopPollingAndSyncAll` and `syncAll`. This is ensure by assigning the producer to the context, before creating the PushManager.
So we can remove the check `if (producer && ...`

NicoZelaya: So why do you also pass the true value in the flagCheck param there then?

EmilianoSanchez: yes. it is not needed.

PR #393, opened 2020-04-14
[Push] Integration tests

EmilianoSanchez:
# JS SDK

## What did you accomplish?

E2E tests for node and browser:
- push-initialization-nopush:
   - testAuthWithPushDisabled
   - testAuthWith401
   - testAuthWithPushDisabled
   - testNoEventSource
   - testNoBase64Support
- push-initialization-retries:
   - testAuthRetries
   - testSSERetries
- push-synchronization
- push-synchronization-retries
- push-fallbacking
- push-refresh-token

NA

## Extra Notes


src/sync/SplitUpdateWorker/index.js
```diff
@@ -18,7 +18,11 @@ export default class SplitUpdateWorker {
   // Private method
   // Preconditions: this.splitProducer.isSynchronizingSplits === false
   __handleSplitUpdateCall() {
-    if (this.maxChangeNumber > this.splitStorage.getChangeNumber()) {
+    // `attempts` is used as an extra stop condition for `__handleSplitUpdateCall` recursion,
+    // to limit at 2 the maximum number of fetches per update event in case `/splitChanges`
+    // requests are failing due to some network or server issue.
+    if (this.maxChangeNumber > this.splitStorage.getChangeNumber() && this.attempts <= 1) {
```
NicoZelaya, line 24: put the one in a "constant" variable. Give it a clear name like max attempts. 

If count starts at 0, but we're limiting to two attempts (including first one and retries) maybe just go with `<` (strictly less) on the condition instead of `<=` (less than) so we can set the value of that to the number of attempts, without thinking much 

_(like it is with the less than, for two attempts, you need to put a 1, basically you need to remember to set it to attempts -1)_

NicoZelaya: Apply this comment to the other workers and any attempts logic.

EmilianoSanchez: no need to apply these changes, since the logic for handling retries has changed to use backoff sequence instead (PR: https://github.com/splitio/javascript-client/pull/397)

EmilianoSanchez: I will clean up this PR, and move these changes into #397

PR #396, opened 2020-04-22
[Push] trigger SDK_UPDATE when split is killed locally

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- handling SDK_UPDATE events on SPLIT_KILL and SPLIT_UPDATE notifications.

- e2e and unit tests updated

## Extra Notes


.travis.yml
```diff
@@ -50,7 +50,7 @@ deploy:
     acl: public_read
     cache_control: "max-age=31536000, public"
     on:
-      branch: development
+      branch: push_sdk_update_on_split_kill
```
NicoZelaya, line 53: Don't forget this baby before merging or after you're done with that canary release.

src/__tests__/browserSuites/push-synchronization-retries.spec.js
```diff
@@ -54,11 +54,11 @@ const MILLIS_DESTROY = 700;
  *  0.2 secs: SPLIT_UPDATE event -> /splitChanges: bad response and then success -> SDK_UPDATE triggered
  *  0.3 secs: SPLIT_UPDATE event with old changeNumber -> SDK_UPDATE not triggered
  *  0.4 secs: MY_SEGMENTS_UPDATE event -> /mySegments/nicolas@split.io: fail and then success -> SDK_UPDATE triggered
- *  0.5 secs: SPLIT_KILL event -> /splitChanges: bad response and network error -> SDK_UPDATE not triggered although cache is updated (killLocally) @TODO review if we should trigger SDK_UPDATE anyway
+ *  0.5 secs: SPLIT_KILL event -> /splitChanges: bad response and network error -> SDK_UPDATE triggered although fetches fail
  *  0.6 secs: MY_SEGMENTS_UPDATE event -> /mySegments/nicolas@split.io: fail twice -> SDK_UPDATE not triggered
  */
 export function testSynchronizationRetries(mock, assert) {
-  assert.plan(14);
+  assert.plan(16);
```
NicoZelaya, line 61: (If you can add a small comment as of why 16, that's great)

src/readiness/index.js
```diff
@@ -55,6 +58,11 @@ function GateContext() {
       setTimeout(() => gate.emit(Events.SDK_READY_FROM_CACHE), 0);
     });
 
+    splits.on(Events.SDK_SPLITS_KILL, () => {
+      // Make it async
+      setTimeout(() => gate.emit(Events.SDK_UPDATE), 0);
```
NicoZelaya, line 63: We would need to ensure this won't fire if the SDK was not ready before by checking that status is SDK_FIRE_UPDATE

src/sync/SplitUpdateWorker/index.js
```diff
@@ -23,26 +25,29 @@ export default class SplitUpdateWorker {
     // requests are failing due to some network or server issue.
     if (this.maxChangeNumber > this.splitStorage.getChangeNumber() && this.attempts <= 1) {
       this.attempts++;
-      this.splitProducer.synchronizeSplits().then(() => {
+      this.splitProducer.synchronizeSplits(this.isSplitKill).then(() => {
         this.__handleSplitUpdateCall();
       });
     } else {
       this.maxChangeNumber = 0;
+      this.isSplitKill = false;
```
NicoZelaya, line 33: Does it need to be an instance property?

src/sync/SplitUpdateWorker/index.js
```diff
@@ -57,8 +62,12 @@ export default class SplitUpdateWorker {
    * @param {string} defaultTreatment default treatment value
    */
   killSplit(changeNumber, splitName, defaultTreatment) {
-    this.splitStorage.killLocally(splitName, defaultTreatment, changeNumber);
-    this.put(changeNumber);
+    this.splitStorage.killLocally(splitName, defaultTreatment, changeNumber).then((updated) => {
+      if (updated) {
```
NicoZelaya, line 66: A favor, please add a big fat note here that, whenever we support an async storage that the SDK itself writes to, we need retries or something in the event of a kill notification.

src/readiness/index.js
```diff
@@ -45,7 +47,8 @@ function GateContext() {
       }
     });
 
-    splits.on(Events.SDK_SPLITS_ARRIVED, () => {
+    splits.on(Events.SDK_SPLITS_ARRIVED, (isSplitKill) => {
+      if(isSplitKill) return;
```
NicoZelaya, line 51: Maybe if you just avoid the assignment of `splitStatus = SPLITS_READY` on the splitKill appearance, you could just use SDK_SPLITS_ARRIVED event, as long as you default that parameter to false (or leave it undefined) and propagate it as true when you emit from the local kill handler.

It'll check the state, if ready hasn't fired nothing is gonna happen. If the SDK is ready to emit updates, it'll emit an update. 

One less event, one less code path, and you avoid the need to propagate the isSplitKill only to be emitted along with the new event.

src/sync/SplitUpdateWorker/index.js
```diff
@@ -57,8 +62,12 @@ export default class SplitUpdateWorker {
...
+        this.splitsEventEmitter.emit(this.splitsEventEmitter.SDK_SPLITS_KILL);
```
NicoZelaya, line 67: After reading my comments on readiness read this:

Here you can emit the SDK_SPLITS_ARRIVED (since actually you've received splits data) but passing the boolean flag as true for readiness to understand it doesn't need to change the status of Splits. 

No need to pass that boolean through the synchronizeSplits + updater path either after that change unless I'm missing something, but seems achievable.

src/__tests__/browserSuites/push-synchronization-retries.spec.js
EmilianoSanchez: this file was moved to PR #397 (comment will be added there)

src/sync/SplitUpdateWorker/index.js
EmilianoSanchez: it was removed

src/readiness/index.js
EmilianoSanchez: removed. I am using SDK_SPLITS_ARRIVED as we talked.

src/readiness/index.js
EmilianoSanchez: great idea

src/sync/SplitUpdateWorker/index.js
EmilianoSanchez: done

PR #422, opened 2020-05-28
SDK_READY[_TIMED_OUT] in consumer mode

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- On consumer mode, SDK_READY event is now emitted once the connection with Redis cache is stablished, instead of being emitted async but immediately when the SDK is created. 
- To avoid a breaking change, the SDK is still operational immediately when created, by setting READY_FROM_CACHE instead of READY in the context (similar as when setting READY_FROM_CACHE in standalone mode using localstorage).
- SDK_READY_TIMED_OUT event is emitted if SDK_READY is not emitted before the time in seconds stablished by `startup.readyTimeout` config param.

- Additionally, `events` and `ioredis` dependencies where moved from optional to required dependencies, since we are not handling the lack of them as we do with `eventsource` optional dependency.

- E2E tests were updated and added to evaluate that readiness events are emitted and ready promise is settled as expected.

## Extra Notes


src/__tests__/node_redis.spec.js
```diff
@@ -118,13 +119,75 @@ tape('NodeJS Redis', function (t) {
...
+          // some asserts to test regular usage
+          assert.equal(await client.getTreatment('UT_Segment_member', 'UT_IN_SEGMENT'), 'on', 'Evaluations using Redis storage should be correct.');
+          assert.equal(await client.getTreatment('other', 'UT_IN_SEGMENT'), 'off', 'Evaluations using Redis storage should be correct.');
+          assert.true(await client.track('nicolas@split.io', 'user', 'test.redis.event', 18), 'If the event was succesfully queued the promise will resolve to true');
+          assert.false(await client.track(), 'If the event was NOT succesfully queued the promise will resolve to false');
+
+          await client.destroy();
+
+          // close server connection
+          server.close().then(assert.end);
```
NicoZelaya, line 184: This doesn't seem to actually wait for the statements on: https://github.com/splitio/javascript-client/pull/422/files#diff-61efb51f5e5ba9bdba3f39bd1fb3fcdcR139-R144
(the two assertions for the SDK yielding the right result once the individual promises resolve)

Should we add meassurements to ensure that those assertions run? using plan or keeping track with a promise all between those and the server.close() ?

src/__tests__/node_redis.spec.js
```diff
@@ -118,13 +119,75 @@ tape('NodeJS Redis', function (t) {
...
+        readyTimestamp = Date.now();
+
+        try {
+          await client.ready();
+          assert.fail('Ready promise keeps being rejected until SDK_READY is emitted');
+        } catch (error) {
+          assert.pass('Ready promise keeps being rejected until SDK_READY is emitted');
+        }
+
+        client.on(client.Event.SDK_READY, async () => {
```
NicoZelaya, line 168: Wouldn't this subscription be better placed outside the server initialization promise callback?

I mean.. does it need to be sequential? It shouldn't matter when the server is created in the whole flow. We should test:
- Timeout when there is no server to connect to, regardless of the readyTimeout value.
- Timeout when there is a server to connect to but the readyTimeout is shorter than the time required to connect. This scenario will end up emitting a SDK_READY shortly after the timeout, and returning a .ready() promise resolved where it was rejected.
- Ready when there is nothing odd.

src/__tests__/node_redis.spec.js
```diff
@@ -133,7 +196,13 @@ tape('NodeJS Redis', function (t) {
 
         client.once(client.Event.SDK_READY_TIMED_OUT, assert.fail);
 
+        const start = Date.now();
         client.once(client.Event.SDK_READY, async () => { // Use SDK_READY event.
+          // ready promise is resolved
```
NicoZelaya, line 201: Can't we do these promise checks in parallel instead of sequentially? After all, that and the event are siblings.

src/storage/node.js
```diff
@@ -25,6 +25,13 @@ const NodeStorageFactory = context => {
       const redis = new RedisAdapter(storage.options);
       const meta = MetaBuilder(settings);
 
+      // subscription to Redis connect event in order to emit SDK_READY
+      const { splits, segments } = context.get(context.constants.READINESS);
+      redis.on('connect', () => {
+        splits.emit(splits.SDK_SPLITS_ARRIVED, false);
+        segments.emit(segments.SDK_SEGMENTS_ARRIVED, false);
```
NicoZelaya, line 32: unused boolean param.

src/index.js
```diff
@@ -35,9 +35,15 @@ export function SplitFactory(config) {
   // We will just log and allow for the SDK to end up throwing an SDK_TIMEOUT event for devs to handle.
   validateApiKey(settings.core.authorizationKey);
 
+  // Put readiness config within context
+  const gateFactory = ReadinessGateFacade();
+  const readiness = gateFactory(settings.startup.readyTimeout);
+  context.put(context.constants.READINESS, readiness);
+  const statusManager = sdkStatusManager(context);
+  context.put(context.constants.STATUS_MANAGER, statusManager);
```
NicoZelaya, line 43: Put a good explanation.

src/__tests__/node_redis.spec.js
EmilianoSanchez: thanks. added the second case by creating a second factory with a very short readyTimedout value

PR #426, opened 2020-06-06
[SDKS-1743] Update API key input validation rules [WIP]

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Updated validation rule for Api keys in standalone mode: they must have more than 20 characters
- [TODO] Updated error message for invalid Api keys
- [TODO] Updated logic for invalid Api keys, as if the SDK is destroyed immediately before being ready, but still emitting SDK_READY_TIMED_OUT. In detail:
  - no requests are done, 
  - minimal internal modules are created, 
  - manager returns empty data since the SDK is never ready, 
  - clients methods behaves as if the SDK is destroyed: `track` returns false, `getTreatment*` return control. Neither events and impressions are tracked, thus: impressionListener and integrations are not invoked, queues are not used, etc.
  - SDK_READY_TIME_OUT is emitted as usual,

- [TODO]

## Extra Notes


EmilianoSanchez:
Delayed. Will be revisited in the future.

PR #431, opened 2020-06-12
[JS size reduction] replace usage of 'for..of' syntax

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Replaced the uses of `for..of` syntax with the following alternatives:
   - `for` syntax
   - `for..in` syntax
   - `Array.prototype.forEach`

It results in removing `forOfIteratorHelper` from the built code, that requires several helpers and ponyfills.

- NA

## Extra Notes


src/producer/updater/SegmentChanges.js
```diff
@@ -46,15 +46,17 @@ export default function SegmentChangesUpdaterFactory(context) {
 
     function segmentsUpdater(segments) {
       const sincePromises = [];
-      for (let segmentName of segments) {
+
+      for (let index = 0; index < segments.length; index++) {
```
NicoZelaya, line 50: Why not i instead of index?

PR #434, opened 2020-06-12
[JS size reduction] remove wrapNativeSuper helper (related to old targets like IE)

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- In old targets such as IE, babel is including the `wrapNativeSuper` helper when extending a class from a global constructor. Since the `Error` global constructor is supported in old environments such as IE 10, we can remove this helper.

NA

## Extra Notes

- [TODO] review caveat with IE10: https://babeljs.io/docs/en/babel-plugin-transform-classes#caveats


EmilianoSanchez:
Won't merge

PR #448, opened 2020-07-06
[JS size reduction] add Map ponyfill

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Added Map polyfill based on popular [ES6 Map Shim](https://www.npmjs.com/package/es6-map), to use if native Map is not available.
- Updated `node/SegmentsCacheInMemory::getRegisteredSegments` to return an array object instead of an iterable (same as the `getRegisteredSegments` method of the other storages)

- Suite tests were run with native Map and polyfilled version.

## Extra Notes


EmilianoSanchez:
Closing PR for the moment. Using refactor of Maps into JS objects instead, but the PR might be revisited in the future.

PR #454, opened 2020-07-24
Fetch specific splits

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Added `SplitFiltersBuilder` module, that validates `splitFilters` config param and parses split filter query string.
- Integrated query string to fetch of split definitions.
- Updated TypeScript declaration file.

- UT for `SplitFiltersBuilder`.
- E2E tests.

## Extra Notes


src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+    log.warn(`Split filters have been configured but will have no effect if mode is not '${STANDALONE_MODE}', since synchronization is being deferred to an external tool`);
+    return;
+  }
+  if (!Array.isArray(splitFilters) || splitFilters.length === 0) {
+    log.warn('splitFilters configuration must be a non-empty array of filters');
+    return;
+  }
+
+  // validate filters and group their values by filter type
+  const filters = {
```
NicoZelaya, line 83: Maybe name it something like "result". Otherwise everything is filter/s

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+    byName: undefined,
+    byPrefix: undefined
+  };
+  const validFilters = splitFilters.filter((filter, index) => {
+    if (filter && isString(filter.type) && Array.isArray(filter.values)) {
+      if (Object.prototype.hasOwnProperty.call(filters, filter.type)) {
+        if (!filters[filter.type]) filters[filter.type] = [];
```
NicoZelaya, line 90: This could be avoided by setting an empty array instead of undefined in line 84 and 85.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+        filters[filter.type] = filters[filter.type].concat(filter.values);
+        return true;
+      } else {
+        log.warn(`'${filter.type}' is an invalid filter. Only 'byName' and 'byPrefix' are valid.`);
+      }
+    } else {
+      log.warn(`Split filter at position '${index}' is invalid. It must be an object with a valid 'type' filter and a list of 'values'.`);
```
NicoZelaya, line 97: Move to input validation too. Maybe instead of position, use the word "index" since that way everyone will know 1 is not the first element on the list.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return false;
+  });
+
+  // remove invalid, deduplicate and order `byName` and `byPrefix` filter values
+  if (filters.byName) filters.byName = filterValidation('byName', filters.byName, 400);
```
NicoZelaya, line 103: byName and byPrefix should come from a constant. So we only need to rename in one place if we ever want to change the wording.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {string} filterName string used for log messages
+ * @param {string[]} list list to validate, deduplicate and sort
+ * @param {number} maxLength
+ * @returns undefined or a sanitized list of values. The list is a non-empty array of unique and alphabetically sorted non-empty strings.
+ * @throws Error if the sanitized list exceeds the length indicated by `maxLength`
+ */
+function filterValidation(filterName, list, maxLength) {
+  // remove invalid values (no strings and empty strings)
+  let result = list.filter(value => {
+    if (!isString(value) || value === '') {
```
NicoZelaya, line 19: You can just check value first, given empty string is falsey.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  // remove duplicated values
+  result = uniq(result);
+
+  // sort values
+  result.sort();
+
+  // check length
+  if (result.length > maxLength) throw new Error(`${maxLength} unique values can be specified at most for '${filterName}' filter. You passed ${result.length}. Please consider reducing the amount or using other filter.`);
```
NicoZelaya, line 33: I owe you this message, ping me after feedback is applied and we'll define it.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+ *  - 'prefixes=<comma-separated-values>': if only byNameList is undefined
+ *  - 'names=<comma-separated-values>&prefixes=<comma-separated-values>': if no one is undefined
+ *
+ * @param {string[] | undefined} byNameList undefined or list of ordered and unique non-empty strings
+ * @param {string[] | undefined} byPrefixList undefined or list of ordered and unique non-empty strings
+ * @returns undefined or string with the `query` component of URL.
+ */
+function queryStringBuilder(byNameList, byPrefixList) {
+  const queryParams = [];
+  if (byNameList && byNameList.length > 0) queryParams.push('names=' + byNameList.map(value => encodeURIComponent(value)).join(','));
```
NicoZelaya, line 56: Don't make it based on the sentences order ideally since it coudl change, as discussed via Slack.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+ *  - '': if both byNameList and byPrefixList are undefined
+ *  - 'names=<comma-separated-values>': if only byPrefixList is undefined
+ *  - 'prefixes=<comma-separated-values>': if only byNameList is undefined
+ *  - 'names=<comma-separated-values>&prefixes=<comma-separated-values>': if no one is undefined
+ *
+ * @param {string[] | undefined} byNameList undefined or list of ordered and unique non-empty strings
+ * @param {string[] | undefined} byPrefixList undefined or list of ordered and unique non-empty strings
+ * @returns undefined or string with the `query` component of URL.
+ */
+function queryStringBuilder(byNameList, byPrefixList) {
```
NicoZelaya, line 54: Method should receive the whole filters set/collection/array/wthv and know how to manage them, to avoid adding an extra param per filter type in the future.

src/utils/settings/splitFilters.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+    } else {
+      log.warn(`Split filter at position '${index}' is invalid. It must be an object with a valid 'type' filter and a list of 'values'.`);
+    }
+    return false;
+  });
+
+  // remove invalid, deduplicate and order `byName` and `byPrefix` filter values
+  if (filters.byName) filters.byName = filterValidation('byName', filters.byName, 400);
+  if (filters.byPrefix) filters.byPrefix = filterValidation('byPrefix', filters.byPrefix, 400);
```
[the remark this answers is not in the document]
EmilianoSanchez: thanks. Good catch!

src/utils/settings/splitFilters.js
EmilianoSanchez: it is done this way to filter also values that are not strings

PR #467, opened 2020-08-13
Data loader [WIP]

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Added support to validate and preload Split data on client-side storage.

- Type declaration tests
- UT validatePreloadedData
- [TODO] UT dataLoader
- [TODO] E2E tests

## Extra Notes


src/storage/DataLoader.js
```diff
@@ -0,0 +1,52 @@
...
+    Object.keys(splitsData).forEach(splitName => {
+      storage.splits.addSplit(splitName, splitsData[splitName]);
+    });
+
+    // add mySegments data
+    let userIdMySegmentsData = preloadedData.mySegmentsData && preloadedData.mySegmentsData[userId];
+    if (!userIdMySegmentsData) {
+      // segmentsData in an object where the property is the segment name and the pertaining value is a stringified object that contains the `added` array of userIds
+      userIdMySegmentsData = Object.keys(segmentsData).filter(segmentName => {
+        const added = JSON.parse(segmentsData[segmentName]).added;
```
NicoZelaya, line 45: Maybe rename added to represent what it stands for here? 
Regardless, it's reaally likely that'll change, since there's no need to respect the BE json after we publish our serializer.

src/storage/DataLoader.js
```diff
@@ -0,0 +1,52 @@
...
+    storage.splits.flush();
+    storage.splits.setChangeNumber(since);
+
+    // splitsData in an object where the property is the split name and the pertaining value is a stringified json of its data
+    Object.keys(splitsData).forEach(splitName => {
+      storage.splits.addSplit(splitName, splitsData[splitName]);
+    });
+
+    // add mySegments data
+    let userIdMySegmentsData = preloadedData.mySegmentsData && preloadedData.mySegmentsData[userId];
```
NicoZelaya, line 41: I think we can remove the userIdMySegmentsData and either just have it `mySegmentsData` (shorter, concise, this has a client scope so we shouldn't need the clarification) or name it 'currentKeyMySegmentsData' or similar.

src/storage/browser.js
```diff
@@ -16,12 +16,13 @@ export const DEFAULT_CACHE_EXPIRATION_IN_MILLIS = 864000000; // 10 days
 const BrowserStorageFactory = context => {
   const settings = context.get(context.constants.SETTINGS);
   const { storage } = settings;
+  let result;
```
NicoZelaya, line 19: Why not "instance" ? It's not the result of an operation.. (it is but it isn't, I mean, it's not a list of values but an instance created by this factory)

src/utils/inputValidation/preloadedData.js
```diff
@@ -0,0 +1,56 @@
+import { isObject, isString } from '../lang';
+import logFactory from '../logger';
+const log = logFactory('', {
+  displayAllErrors: true
```
NicoZelaya, line 4: I'm not so sure about this... Why is it true here? What's the reasoning behind it?

src/storage/DataLoader.js
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Storage-agnostic adaptation of `loadDataIntoLocalStorage` function
+   * (https://github.com/godaddy/split-javascript-data-loader/blob/master/src/load-data.js)
+   *
+   * @param {Object} storage storage for client-side
+   * @param {Object} userId main user key defined at the SDK config
+   */
+  return function loadData(storage, userId) {
+    // Do not load data if current preloadedData is empty
+    if (Object.keys(preloadedData).length === 0) {
```
NicoZelaya, line 19: Be consistent with the 1 line if's on inputvalidation and remove the brackets. Or add them there, but choose one.

src/storage/DataLoader.js
```diff
@@ -0,0 +1,52 @@
...
+      return;
+    }
+
+    const { segmentsData = {}, since = 0, splitsData = {} } = preloadedData;
+
+    const currentSince = storage.splits.getChangeNumber();
+
+    // Do not load data if current localStorage data is more recent
+    if (since <= currentSince) {
```
NicoZelaya, line 28: One liner if, standarize.

src/utils/inputValidation/preloadedData.js
```diff
@@ -0,0 +1,56 @@
...
+});
+
+function validateSinceData(maybeSince, method) {
+  if (maybeSince > -1) return true;
```
NicoZelaya, line 8: Be careful. If you do `maybeSince > -1` with undefined it yields false, but if you do it with null it yields true. The validation should check that's a number before anything, then you see how big it is. 

Null is probably what you'll get if the BE serializer can't assign a proper one or there's some sort of parsing error somewhere (remember the data serializer won't be running in this same app), given that most languages won't send undefined.

src/utils/inputValidation/preloadedData.js
```diff
@@ -0,0 +1,56 @@
...
+  log.error(`${method}: preloadedData.since must be a positive number.`);
+  return false;
+}
+
+function validateSplitsData(maybeSplitsData, method) {
+  if (isObject(maybeSplitsData)) {
+    const splitNames = Object.keys(maybeSplitsData);
+    if (splitNames.length > 0 && splitNames.every(splitName => isString(maybeSplitsData[splitName]))) return true;
+  }
+  log.error(`${method}: preloadedData.splitsData must be a map of split names to their serialized definitions.`);
```
NicoZelaya, line 18: Maybe we should make this a separate validation. 
For an empty list we throw a warning (since it's possible to have an empty list of splits for an env).

For a list with items, if not all have the right format (keep isString for now or go and add the basic validation for Splits already, to check the key parts of the JSON) we'll log an error for invalid values.

src/utils/inputValidation/preloadedData.js
```diff
@@ -0,0 +1,56 @@
...
+  log.error(`${method}: preloadedData.mySegmentsData must be a map of user keys to their list of segment names.`);
+  return false;
+}
+
+function validateSegmentsData(maybeSegmentsData, method) {
+  if (isObject(maybeSegmentsData)) {
+    const segmentNames = Object.keys(maybeSegmentsData);
+    if (segmentNames.length > 0 && segmentNames.every(segmentName => isString(maybeSegmentsData[segmentName]))) return true;
+  }
+  log.error(`${method}: preloadedData.segmentsData must be a map of segment names to their serialized definitions.`);
```
NicoZelaya, line 40: Same, if list is empty it shouldn't be an error.

src/utils/inputValidation/preloadedData.js
```diff
@@ -0,0 +1,56 @@
...
+function validateMySegmentsData(maybeMySegmentsData, method) {
+  if (isObject(maybeMySegmentsData)) {
+    const userKeys = Object.keys(maybeMySegmentsData);
+    if (userKeys.length > 0 && userKeys.every(userKey => {
+      const segmentNames = maybeMySegmentsData[userKey];
+      // an empty list is valid
+      return Array.isArray(segmentNames) && segmentNames.every(segmentName => isString(segmentName));
+    })) return true;
+  }
+  log.error(`${method}: preloadedData.mySegmentsData must be a map of user keys to their list of segment names.`);
```
NicoZelaya, line 31: If list is empty it shouldn't be an error.

src/utils/logger/index.js
```diff
@@ -41,7 +41,7 @@ const initialState = String(
       localStorage.getItem(LS_KEY) : ''
 );
 
-const createLog = (namespace, options = {}) => new Logger(namespace, merge(options, defaultOptions));
+const createLog = (namespace, options = {}) => new Logger(namespace, merge(defaultOptions, options));
```
NicoZelaya, line 44: Ain't this overriding the default options? We are assigning to a new variable in the `merge` method but objects are assigned per reference, not a copy. 

Are you sure of this? The fact that all errors were hidden and the show level is shown is enforced on purpose. We want a centralized knob to tweak ALL loggers.

src/utils/inputValidation/preloadedData.js
EmilianoSanchez: Thanks. It is a good idea to use the existing split validation utils

src/storage/browser.js
EmilianoSanchez: definitively right

src/utils/logger/index.js
EmilianoSanchez: you are right. I will rollback this change, and recheck it in another moment.

PR #487, opened 2020-09-16
added support for beacon/count

mmelograno:
# JS SDK

## Extra Notes


src/listeners/browser.js
```diff
@@ -70,6 +78,16 @@ export default class BrowserSignalListener {
     }
   }
 
+  _flushImpressionsCount() {
+    const pf = fromImpressionsCountCollector(this.impressionsCounter);
```
NicoZelaya, line 82: Name should be something with payload to match the others ideally. it should represent it is the same payload that we send to the other endpoint.

NicoZelaya (approved):

EmilianoSanchez (approved):

PR #495, opened 2020-11-03
Fix import issue for webpack 5

andrew-t-james-hm:
# JS SDK

## What did you accomplish?
resolve issue that blocks webpack 5 from being able to complete a build.
```Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)```
Nonbreaking change, all existing tests should be passing.
## Extra Notes


NicoZelaya:
Thank you very much for your contribution @andrew-t-james-hm ! We'll take it from here and release in an upcoming patch. Will ping you here once it's officially out but should be by EOW at most.

andrew-t-james-hm:
@NicoZelaya Thanks for the quick response, much appreciated.

PR #510, opened 2021-02-09
Type definitions update

EmilianoSanchez:
# JS SDK

## What did you accomplish?

Updated EventEmitter type definition to not require @types/node, in order to avoid possible conflicts with the "dom" type definitions of build-in APIs such as setTimeout or console.

Declaration type tests were extended to assert that the EventEmitter interface is type compatible with the NodeJS.EventEmitter interface on different versions of @types/node.

## Extra Notes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ deploy:
     acl: public_read
     cache_control: "max-age=31536000, public"
     on:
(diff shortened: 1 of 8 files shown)


NicoZelaya (approved):
Thank you doc!

PR #540, opened 2021-05-19
throw error instead of string

scottbessler:
for ref: https://nodejs.dev/learn/error-handling-in-nodejs

# JS SDK

## What did you accomplish?

It is preferable in javascript, and especially in node to throw Error objects. I noticed that SplitError already existed, so wrapped the thrown strings in that.

## Extra Notes

Unfortunately this is technically backwards incompatible as anyone catching the error will get an error object now.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ import keyParser from './utils/key/parser';
 import logFactory, { API } from './utils/logger';
 const log = logFactory('splitio');
 import tracker from './utils/timeTracker';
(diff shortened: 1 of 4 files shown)


EmilianoSanchez:
thank you @scottbessler  for the comments.

We had started to consider this good practice in other of our Javascript repositories, but not in this one yet. We will merge your PR to include it in the coming patch release, together with a linter rule to force the recommendation. 

Best,
Emiliano

scottbessler:
🎉

PR #601, opened 2021-12-02
attributes in memory storage

emmaz90:
# JS SDK

## What did you accomplish?
Create attribute in memory storage
## Extra Notes

### DIFF
--- a/src/storage/AttributesCache/InMemory.js
+++ b/src/storage/AttributesCache/InMemory.js
@@ -1,0 +1,78 @@
+import objectAssign from 'object-assign';
+
+class AttributesCacheInMemory {
+
+  constructor() {
+    this.attributesCache = {};
+  }
+
+  /**
+   * Create or update the value for the given attribute
+   * 
+   * @param {string} attributeName attribute name
+   * @param {Object} attributeValue attribute value
+   * @returns {boolean} the attribute was stored 
+   */
+  setAttribute(attributeName, attributeValue) {
+    this.attributesCache[attributeName] = attributeValue;
+    return true;
+  }
+
+  /**
+   * Retrieves the value of a given attribute
+   * 
+   * @param {string} attributeName attribute name
+   * @returns {Object?} stored attribute value
+   */
+  getAttribute(attributeName) {
+    return this.attributesCache[attributeName];
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


src/storage/__tests__/AttributesCache/InMemory.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+  assert.true(cache.getAttribute('attributeName2') === 'attributeValue2');
+  assert.true(cache.getAttribute('attributeName3') === 'attributeValue3');
+  assert.true(cache.getAttribute('attributeName4') === 'attributeValue4');
+
+  cache.clear();
+
+  values = cache.getAll();
+
+  assert.true(values['attributeName2'] === undefined);
```
mmelograno, line 34: assert.equal(Object.keys(values).length, 0, 'It should be zero after clearing attributes') instead

src/storage/AttributesCache/InMemory.js
```diff
@@ -0,0 +1,78 @@
...
+   * Removes a given attribute from the map
+   * 
+   * @param {string} attributeName attribute to remove
+   * @returns {boolean} attribute removed
+   */
+  removeAttribute(attributeName) {
+    if (this.attributesCache[attributeName]) {
+      delete this.attributesCache[attributeName];
+      return true;
+    } else {
```
mmelograno, line 63: this else can be removed and just use the return false

src/storage/__tests__/AttributesCache/InMemory.spec.js
```diff
@@ -0,0 +1,39 @@
...
+tape('ATTRIBUTE CACHE / In Memory', assert => {
+  const cache = new AttributesCacheInMemory();
+
+  cache.setAttribute('attributeName1', 'attributeValue1');
+  cache.setAttribute('attributeName2', 'attributeValue2');
+
+
+  let values = cache.getAll();
+
+  assert.true(values['attributeName1'] === 'attributeValue1');
```
mmelograno, line 13: maybe replace assert.true for assert.equal ?

mmelograno (changes requested):

mmelograno (approved):

PR #602, opened 2021-12-02
Attributes binding api in splitclient

emmaz90:
# JS SDK

## What did you accomplish?
Bind attribute storage api to browser split client

## Extra Notes

### DIFF
--- a/src/client/__tests__/attributesDecoration.spec.js
+++ b/src/client/__tests__/attributesDecoration.spec.js
@@ -23,22 +23,20 @@ tape('ATTRIBUTES DECORATION / storage', assert => {
 
   let values = client.getAttributes();
 
-  assert.equal(values['attributeName1'], 'attributeValue1', 'It should be equal');
-  assert.equal(values['attributeName2'], 'attributeValue2', 'It should be equal');
+  assert.deepEqual(values, { attributeName1: 'attributeValue1', attributeName2: 'attributeValue2' }, 'It should be equal');
 
   client.removeAttribute('attributeName1');
+  client.setAttribute('attributeName2', 'newAttributeValue2');
 
   assert.equal(client.getAttribute('attributeName1'), undefined, 'It should throw undefined');
-  assert.equal(client.getAttribute('attributeName2'), 'attributeValue2', 'It should be equal');
+  assert.equal(client.getAttribute('attributeName2'), 'newAttributeValue2', 'It should be equal');
 
   client.setAttributes({
     'attributeName3': 'attributeValue3',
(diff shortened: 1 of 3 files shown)


src/client/__tests__/attributesDecoration.spec.js
```diff
@@ -0,0 +1,50 @@
...
+tape('ATTRIBUTES DECORATION / storage', assert => {
+
+  const client = AttributesDecorationMockedClient({});
+
+  client.setAttribute('attributeName1', 'attributeValue1');
+  client.setAttribute('attributeName2', 'attributeValue2');
+
+  let values = client.getAttributes();
+
+  assert.equal(values['attributeName1'], 'attributeValue1', 'It should be equal');
```
EmilianoSanchez, line 26: You can use `deepEqual` here:
```
assert.deepEqual(values, { attributeName1: 'attributeValue1', attributeName2: 'attributeValue2' }, 'It should be equal');
```

src/client/__tests__/attributesDecoration.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  assert.equal(values['attributeName2'], 'attributeValue2', 'It should be equal');
+
+  client.removeAttribute('attributeName1');
+
+  assert.equal(client.getAttribute('attributeName1'), undefined, 'It should throw undefined');
+  assert.equal(client.getAttribute('attributeName2'), 'attributeValue2', 'It should be equal');
+
+  client.setAttributes({
```
EmilianoSanchez, line 34: It would be nice to update `attributeName2` here with a different value that the initial one, to assert that an already existing attribute is properly updated.

src/client/__tests__/attributesDecoration.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+  assert.equal(values['attributeName2'], 'attributeValue2', 'It should be equal');
+  assert.equal(values['attributeName3'], 'attributeValue3', 'It should be equal');
+  assert.equal(values['attributeName4'], 'attributeValue4', 'It should be equal');
+
+  client.clearAttributes();
+
+  values = client.getAttributes();
+
+  assert.equal(Object.keys(values).length, 0, 'It should be zero after clearing attributes');
```
EmilianoSanchez, line 47: maybe `assert.deepEqual(values, {}, 'Attributes should be an empty object')` ? Anyway, it is just a matter of preference.

EmilianoSanchez (approved):
LGTM Emma. Just a few comments regarding unit tests, that would be nice to apply.

Issue #621, opened 2022-01-25
Cannot read properties of undefined (reading 'response')


amakhrov:
Thrown by the error handling code inside `.catch`: https://github.com/splitio/javascript-client/blob/development/src/services/transport/index.js#L22

I don't know the specific steps to reproduce, but it happens to a bunch of our users. In general, one cannot assume a caught error has any particular data type (or even is defined)

EmilianoSanchez:
Thanks @amakhrov  for  #622. 
That PR is something we will include in next release for sure. As you mentioned, it should not assume that all Fetch API implementations will reject with a particular error type.

PR #622, opened 2022-01-25
Fix a null error in transport/index.js

amakhrov:
# JS SDK

## What did you accomplish?
Fixes https://github.com/splitio/javascript-client/issues/621

Good question :) . I don't have clear reproduction for the issue.


EmilianoSanchez (approved):
LGTM. No need to add new tests

PR #643, opened 2022-02-25
Localstorage key scoped to prefix

NicoZelaya:
# JS SDK

## What did you accomplish?
- Scoped my segments data in localstorage inside the same prefix than the other keys.

- Run the unit tests
- Validate running on a web, you can introduce a few old keys from the console. Reach out to me if you want some example sentences.

## Extra Notes
- The goal is to make the change without leaving old cache keys behind.


EmilianoSanchez (approved):
LGTM

PR #651, opened 2022-03-03
Redis intransit encryption support

dirkonly:
# JS SDK

## What did you accomplish?

in-transit encryption support added. The 'tls' option was not sent to the redis client and therefor connections to our encrypted redis could not be made

Initialize splitio in consumer mode and use an encrypted redis storage


EmilianoSanchez:
Hi, @dirkonly , 

We are working on a refactor of the SDK internal dependencies, to reuse code from https://github.com/splitio/javascript-commons .
Therefore, we will not merge the PR here, but we will use it as a reference to implement the Redis encryption support in the javascript-commons repository.

Even though we won't merge this PR, we're thankful for the contribution.

EmilianoSanchez:
Hi @dirkonly ,

I am closing your PR, since we have applied it to these files: [RedisAdapter](https://github.com/splitio/javascript-commons/blob/development/src/storages/inRedis/RedisAdapter.ts), [storage/node](https://github.com/splitio/javascript-client/blob/development/src/settings/storage/node.js), recently released with version 10.18.0 of the SDK.

Thank you, 
Emiliano

PR #653, opened 2022-03-16
[JS-commons migration] Redis intransit encryption support

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Added support for TLS config options for Redis. Based on #651 
- Updated type definitions to describe storage options for Redis

- Unit tests

## Extra Notes

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -46,7 +46,7 @@ jobs:
         run: BUILD_BRANCH=$(echo "${GITHUB_REF#refs/heads/}") BUILD_COMMIT=${{ github.sha }} npm run build
 
       - name: Deploy S3 Development
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/redis_intransit_encryption_support' }}
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/development' }}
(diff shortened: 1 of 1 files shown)


src/settings/storage/node.js
```diff
@@ -30,7 +30,12 @@ export function validateStorage(settings) {
...
         tls,
         connectionTimeout,
-        operationTimeout
+        operationTimeout,
+        redisTlsEnabled,
+        redisTlsCaCertificates,
+        redisTlsSkipNameValidation,
+        redisTlsClientCertificate,
+        redisTlsClientKey
       } = options;
```
gsiwatch, line 34: Not a part of this PR: but just wondering
will it be possible to define a type for this Object so that clients using it can know what options they can use instead of coming here and checking the code?

Currently this is what I am seeing in Node SDK 10.17.0
```
  readonly storage: {
    prefix: string,
    options: Object,  <-- type here
    type: StorageType
  },

emmaz90 (approved):

NicoZelaya (approved):

src/settings/storage/node.js
EmilianoSanchez: That's a good point. We will include the type of the `options` object :)

PR #675, opened 2022-05-23
EventSource dependency update [WIP]

EmilianoSanchez:
# JS SDK

## Extra Notes


EmilianoSanchez:
We will not update EventSource dependency to v2 for now, because it implies dropping streaming support for NodeJS below v12.

PR #711, opened 2022-08-29
[SDKS-5970] Add refresh rate property to bloom filter

emmaz90:
# JS SDK

## What did you accomplish?
Add refreshRate property to bloom filter so it will have a cache cleaner every 24 hs

## Extra Notes


EmilianoSanchez (approved):
LGTM

PR #714, opened 2022-09-09
[SDKS-6093] Add e2e tests for redis optimized and none strategy

emmaz90:
# JS SDK

## What did you accomplish?
Added E2E tests for optimized and none strategy in redis
Fix tests for not counting first impression

## Extra Notes

### DIFF
--- a/src/__tests__/consumer/node_redis.spec.js
+++ b/src/__tests__/consumer/node_redis.spec.js
@@ -296,7 +296,7 @@ tape('NodeJS Redis', function (t) {
           {'f':'hierarchical_splits_testing_off','ks':['UT_Segment_member']},
           {'f':'hierarchical_splits_testing_on_negated','ks':['UT_Segment_member']},
         ];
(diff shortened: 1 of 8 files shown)


src/__tests__/consumer/node_redis.spec.js
```diff
@@ -146,6 +148,242 @@ tape('NodeJS Redis', function (t) {
...
+      });
+  });
+  
+  t.test('Regular usage - NONE strategy', assert => {
+    config.sync.impressionsMode = NONE;
+    initializeRedisServer()
+      .then(async (server) => {
+        const expectedConfig = '{"color":"brown"}';
+        const timeFrame = Date.now();
+        const expectedImpressionCount = [
```
EmilianoSanchez, line 277: Nice work Emma!
Just a single comment.

If `expectedImpressionCount` is the same for the NONE and the OPTIMIZED tests, we can move it to line 71 and reuse in both tests. Same with `expectedUniqueKeys` and `expectedConfig` (this one is used 3 times)

EmilianoSanchez (approved):

PR #762, opened 2023-09-28
[SDKS-7563] Stabilize integration tests

emmaz90:
# JS SDK

## What did you accomplish?
Stabilize Integration tests for flag sets implementation

## Extra Notes

### DIFF
--- a/src/__tests__/browserSuites/ready-from-cache.spec.js
+++ b/src/__tests__/browserSuites/ready-from-cache.spec.js
@@ -477,7 +477,7 @@ export default function (fetchMock, assert) {
       events: 'https://events.baseurl/readyFromCache_5'
     };
     localStorage.clear();
-    t.plan(7);
+    t.plan(6);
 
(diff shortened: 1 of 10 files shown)


src/__tests__/browserSuites/ready-from-cache.spec.js
```diff
@@ -503,10 +503,6 @@ export default function (fetchMock, assert) {
     const client = splitio.client();
     const manager = splitio.manager();
 
-    client.once(client.Event.SDK_READY_FROM_CACHE, () => {
```
EmilianoSanchez, line 506: Rather than simply removing asserts, it is better to update them to keep information of the expected behaviour:

```
client.once(client.Event.SDK_READY_FROM_CACHE, () => {
  t.fail('It should not emit SDK_READY_FROM_CACHE because ...');
  t.end();
});
```

Idem for the other removed asserts.

EmilianoSanchez (changes requested):

src/__tests__/browserSuites/ready-from-cache.spec.js
EmilianoSanchez: With this change, none of the tests below line 472 (/** Fetch specific splits **/) is asserting that the SDK_READY_FROM_CACHE event is emitted.

So, I would create a new test for the following basic case: 
```
  assert.test(t => { // Testing when we start with cached data with split filter, and the same split filter is provided in the config
```
This should emit the event because the config has the same split filter as the one in the cache.

EmilianoSanchez (approved):

PR #814, opened 2024-07-24
Large Segments tests: polling mode

EmilianoSanchez:
# JS SDK

## What did you accomplish?

- Fix some typos and rename `fallbacking` by `fallback`.
- Update `push-fallback` tests to validate that `myLargeSegments` are fetched with polling when streaming mode switches to polling mode. 

## Extra Notes


javrudsky (approved):
👍

PR #897, opened 2025-10-24
[FME-10568] update js commons and add fallback treatment e2e tests

ZamoraEmmanuel:
# JS SDK

## Extra Notes


EmilianoSanchez (approved):

EmilianoSanchez (changes requested):
Emma, we need to check that TypeScript definitions are correct and include the new `fallbackTreatments` configuration options for the different SDK modes, by updating the `ts-tests/index.ts` file with the corresponding tests.

